PR #30, opened 2016-03-25
cookie cutter with some templating + pre and post gen script hooks

wismer:


jschneier:
Closing in favor of #32

PR #31, opened 2016-03-25
Dot env setup

wismer:
based on #30

PR #32, opened 2016-03-25
Contrib site migration

wismer:
based on #31

### DIFF
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -1,17 +1,17 @@
 {
     "project_name": "testapp",
     "repo_name": "{{ cookiecutter.project_name|replace(' ', '_') }}",
-    "djoser": ["basic", "with pw reset", "all"],
+    "djoser": ["basic", "all"],
     "client_repo_name": "{{ cookiecutter.repo_name}}-client",
     "django_registration": ["y", "n"],
     "stripe": ["y", "n"],
-    "stripe_key": "lk;gjsglkjsdlgkjsdlkgfj",
     "aws_access_key_id": "",
     "aws_storage_secret": "",
+    "org_name": "Lightmatter",
     "description": "A short description of the project.",
     "domain_name": "example.com",
     "version": "0.1.0",
     "timezone": "UTC",
     "now": "2016/01/07",
-    "year": "{{ cookiecutter.now[:4] }}",
+    "year": "{{ cookiecutter.now[:4] }}"
 }
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -1,12 +1,17 @@
 {
     "project_name": "testapp",
     "repo_name": "{{ cookiecutter.project_name|replace(' ', '_') }}",
-    "author_name": "Your Name",
-    "email": "Your email",
+    "djoser": ["basic", "with pw reset", "all"],
+    "client_repo_name": "{{ cookiecutter.repo_name}}-client",
+    "django_registration": ["y", "n"],
+    "stripe": ["y", "n"],
+    "stripe_key": "lk;gjsglkjsdlgkjsdlkgfj",
+    "aws_access_key_id": "",
+    "aws_storage_secret": "",
     "description": "A short description of the project.",
     "domain_name": "example.com",
     "version": "0.1.0",
     "timezone": "UTC",
     "now": "2016/01/07",
-    "year": "{{ cookiecutter.now[:4] }}"
+    "year": "{{ cookiecutter.now[:4] }}",
 }
--- a/hooks/post_gen_project.sh
+++ b/hooks/post_gen_project.sh
@@ -52,14 +52,15 @@ EOF
 
 python manage.py migrate
 
-#link up with git!
-if [ -d .git ]; then
-  echo "Git exists";
-else
-    echo "Setting up Git"
-    git init .
-    git remote add origin "git@github.com:Lightmatter/{{ cookiecutter.repo_name }}.git"
-    #todo - add all and make initial push
-fi
+
+echo "Setting up Git"
+git init .
+git remote add origin "git@github.com:{{cookiecutter.org_name}}/{{ cookiecutter.repo_name }}.git"
+git flow init -d
+git add .
+git commit -m "initial commit"
+git push
 
 chmod +x manage.py
+
+mv .env.example .env
--- a/hooks/post_gen_project.sh
+++ b/hooks/post_gen_project.sh
@@ -1,0 +1,65 @@
+#!/bin/bash
+
+source `which virtualenvwrapper.sh`
+unset PYTHONDONTWRITEBYTECODE
+
+ENV_NAME="{{ cookiecutter.repo_name }}"
+ENV_PATH="${PROJECT_HOME}/${ENV_NAME}"
+
+workon $ENV_NAME
+cd $ENV_PATH
+
+if [ ! -d ${HOME}/.pip-packages ]
+then
+    mkdir -p ${HOME}/.pip-packages
+fi
+
+
+
+if [  -d $WORKON_HOME/{{ cookiecutter.repo_name }}/build/ ]
+
+then
+    rm -rf $WORKON_HOME/{{ cookiecutter.repo_name }}/build/
+fi
+
+
+pip install pip-accel
+which pip-accel
+if [ $? -ne 0 ]; then
+    pip install --download ${HOME}/.pip-packages --exists-action w -r requirements-dev.txt
+    pip install --no-index --exists-action w --find-links=file://${HOME}/.pip-packages/ -r requirements-dev.txt
+else
+    pip install -r requirements-dev.txt
+fi
+
+RESULT=`psql -l | grep "{{ cookiecutter.repo_name }}" | wc -l | awk '{print $1}'`;
+if test $RESULT -eq 0; then
+    echo "Creating Database";
+    psql -c "create role {{ cookiecutter.repo_name }} with createdb encrypted password '{{ cookiecutter.repo_name }}' login;"
+    psql -c "alter user {{ cookiecutter.repo_name }} superuser;"
+    psql -c "create database {{ cookiecutter.repo_name }} with owner {{ cookiecutter.repo_name }};"
+else
+    echo "Database exists"
+fi
+
+export DJANGO_SETTINGS_MODULE=$ENV_NAME.$ENV_NAME.settings.local
+
+SECRET_KEY=`python -c 'import random; print("".join([random.choice("abcdefghijklmnopqrstuvwxyz0123456789!@#$%^&*(-_=+)") for i in range(50)]))'`
+#todo - git flow init
+cat >> .env <<EOF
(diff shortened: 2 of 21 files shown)


cookiecutter.json
```diff
@@ -1,12 +1,17 @@
...
+    "stripe_key": "lk;gjsglkjsdlgkjsdlkgfj",
+    "aws_access_key_id": "",
+    "aws_storage_secret": "",
     "description": "A short description of the project.",
     "domain_name": "example.com",
     "version": "0.1.0",
     "timezone": "UTC",
     "now": "2016/01/07",
-    "year": "{{ cookiecutter.now[:4] }}"
+    "year": "{{ cookiecutter.now[:4] }}",
```
jschneier: JSON does not permit a trailing comma.

{{cookiecutter.repo_name}}/.env
```diff
@@ -0,0 +1,15 @@
+
```
jschneier: Add a `.env.example` file and then add `.env` to `.gitignore`.

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/settings/local.py
```diff
@@ -23,7 +23,7 @@
     'default': {
```
jschneier: So now that we have this environ stuff going which is great! But that means that basically all of our environment variable settings can be moved to `base.py`. And the `local.py` or whatever we are calling it can then just have stuff like backends and template options.

In particular, `SECRET_KEY` and `STRIPE_*_KEY` should be moved to `base.py` and `STRIPE_*` should be behind the cookiecutter flag.

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/settings/base.py
```diff
@@ -1,8 +1,14 @@
 # Django settings for project project.
-from os import environ
-from django.core.exceptions import ImproperlyConfigured
+from environ import Env, Path
 import pathlib
+from django.core.exceptions import ImproperlyConfigured
+
+env = Env()
```
jschneier: I don't think we are putting `.env` on heroku so maybe this call should just be in `local.py`? Need to think about the organization some.

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/account/urls.py
```diff
@@ -1,4 +1,22 @@
...
+    url(r'^logout/', LogoutView.as_view(), name='user-logout'),
+{%- if cookiecutter.djoser != 'basic' -%}
+    url(r'^register/user/', RegistrationView.as_view(), name='user-register'),
+    url(r'^password/reset/$', PasswordResetView.as_view(), name='password-reset'),
+    url(r'^password/reset/confirm/$', PasswordResetConfirmView.as_view(), name='password-confirm'),
+{%- endif -%}
+]
+
+urlpatterns = [
+    url(r'^api/auth/', include(authurls))
```
jschneier: I think we want to do this prefixing at the `{{ cookiecutter.repo_name }}/urls.py` level.

cookiecutter.json
```diff
@@ -1,12 +1,17 @@
 {
     "project_name": "testapp",
     "repo_name": "{{ cookiecutter.project_name|replace(' ', '_') }}",
-    "author_name": "Your Name",
-    "email": "Your email",
+    "djoser": ["basic", "with pw reset", "all"],
```
jschneier: Is `pw reset` used?

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/account/urls.py
wismer: @jschneier wouldn't we want to the auth stuff to be at the account level? If I do it at the lower level, how do I add new url paths with `include('{{ cookiecutter.repo_name }}.account.urls')`? (assuming that's the way to grab it)

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/settings/base.py
gone: It will just throw a warning if the file is missing, so the calls can be the same local vs prod

hooks/post_gen_project.sh
```diff
@@ -52,14 +52,13 @@ EOF
...
-    git init .
-    git remote add origin "git@github.com:Lightmatter/{{ cookiecutter.repo_name }}.git"
-    #todo - add all and make initial push
-fi
 
-chmod +x manage.py
+echo "Setting up Git"
+git init .
+git remote add origin "git@github.com:{{cookiecutter.org_name}}/{{ cookiecutter.repo_name }}.git"
+git flow init
```
gone: I think init needs some args to make it work w/o input

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/settings/base.py
```diff
@@ -1,9 +1,15 @@
...
 
 
+Env.read_env('.env')
+DEBUG = env('DEBUG')
```
jschneier: I'd prefer if `DEBUG` is not managed by env var. It's too dangerous. Let's set it to `False` in the base and then `True` in local.py only.

I think we can say the same about `SSLIFY`, not sure.

{{cookiecutter.repo_name}}/.env
jschneier: Cool now just remove this file

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/settings/local.py
```diff
@@ -68,7 +68,4 @@
 )
 
 TEMPLATES[0]['OPTIONS']['string_if_invalid'] = 'BAD TEMPLATE VARIABLE: %s'
-SECRET_KEY = get_env_setting("DJANGO_SECRET_KEY", default='CHANGEME!!!')
-
-STRIPE_PUBLIC_KEY = ''
-STRIPE_SECRET_KEY = ''
+SECRET_KEY = env('SECRET_KEY')
```
jschneier: This one _is_ added to the `.env` so I think it can be in the base.py

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/settings/heroku.py
```diff
@@ -9,8 +9,11 @@
 DATABASES = {}
 DATABASES['default'] = dj_database_url.config()
 
+# Honor the 'X-Forwarded-Proto' header for request.is_secure()
 SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
-ALLOWED_HOSTS = get_env_setting('ALLOWED_HOSTS').split()
+
+ALLOWED_HOSTS = []  # TODO
```
jschneier: Couldn't get this to work??

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/account/urls.py
jschneier: Yep, you're correct.

hooks/post_gen_project.sh
```diff
@@ -0,0 +1,64 @@
...
+SECRET_KEY="${SECRET_KEY}"
+EOF
+
+python manage.py migrate
+
+
+echo "Setting up Git"
+git init .
+git remote add origin "git@github.com:{{cookiecutter.org_name}}/{{ cookiecutter.repo_name }}.git"
+git flow init
```
jschneier: I think we're still missing the flag to do `git flow` automatically, `-d`.

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/settings/base.py
```diff
@@ -1,8 +1,12 @@
 # Django settings for project project.
-from os import environ
-from django.core.exceptions import ImproperlyConfigured
+from environ import Env, Path
 import pathlib
+from django.core.exceptions import ImproperlyConfigured
 
+env = Env()
```
jschneier: What happened to `.env`? I'm not saying it's wrong I'm just wondering. I looked at the source and it does some funky things to trace it back.

{{cookiecutter.repo_name}}/.env
jschneier: @wismer can you delete the `.env` file? You might need to update the init script to `mv env.example .env`

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/settings/heroku.py
```diff
@@ -9,11 +9,13 @@
 DATABASES = {}
 DATABASES['default'] = dj_database_url.config()
 
+# Honor the 'X-Forwarded-Proto' header for request.is_secure()
 SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
-ALLOWED_HOSTS = get_env_setting('ALLOWED_HOSTS').split()
-STATIC_ROOT = str(PROJECT_ROOT / 'static')
 
-redis_url = urlparse(os.environ.get('REDISTOGO_URL', 'redis://localhost:6959'))
+ALLOWED_HOSTS = []  # TODO
```
jschneier: Switch this to `env('ALLOWED_HOSTS').split('|')`

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/settings/base.py
wismer: @jschneier By default it looks at `.env` and if it can't find it, I believe it pulls from `ENV`. I tried it out in the marex-dev shell and it pulls out all the heroku env variables without having a `.env` file.

jschneier: Yeah, at heart it just wraps `os.envrion` and optionally reads in a file. I should have put this a line lower, I meant in the `.read_env` call.

PR #33, opened 2016-04-07
Mailhog

wismer:
this will need to be rebased eventually after #31 gets merged. All the mailhog stuff is in the last commit.


{{cookiecutter.repo_name}}/.env
```diff
@@ -0,0 +1,15 @@
+
+{% if cookiecutter.stripe %}
+STRIPE_SECRET_KEY="{{ cookiecutter.stripe_key }}"
+STRIPE_PUBLIC_KEY=""
+{%- endif %}
+
+AWS_ACCESS_KEY_ID=""
```
gone: Is .env only for local settings? Normally locally we do everything on the harddrive and not remotely

gone:
Other than possibly removing the aws settings for local development I do'nt see anything else I'd change

PR #38, opened 2016-07-11
add apps.py files to account, app and util and specify default_app_co…

jake-kent:
…nfig in each app's __init__.py


{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/app/apps.py
```diff
@@ -0,0 +1,6 @@
+from django.apps import AppConfig
+
+class MainAppConfig(AppConfig):
```
jschneier: I want to de-emphasize (and probably eventually remove) ``app``. You can call this `AppConfig` so long as your alias the import:
```python
from django.apps import AppConfig as DjangoAppConfig

class AppConfig(DjangoAppConfig):
  ...
```

I actually think we should rename this to `home` pursuant to @gone suggestion.

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/__init__.py
```diff
@@ -1 +1 @@
-
+default_app_config = 'app.apps.MainAppConfig'
```
jschneier: I think this needs to be `{{ cookiecutter.repo_name }}.app.apps.MainAppConfig` but confirmation would be nice.

jake-kent: Already within the first {{ cookiecutter.repo_name }} folder, so not necessary.

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/home/apps.py
```diff
@@ -0,0 +1,6 @@
+from django.apps import AppConfig as DjangoAppConfig
+
+class HomeConfig(DjangoAppConfig):
+	name = "main_app"
```
gone: Should be home

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/home/management/commands/loadcsv.py
```diff
@@ -1,6 +1,6 @@
 from django.core.management.base import BaseCommand, CommandError
 
-#from app.models import SiteExperience
+#from home.models import SiteExperience
```
gone: pull this line out

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/__init__.py
```diff
@@ -1 +1 @@
-
+default_app_config = 'home.apps.MainAppConfig'
```
gone: I think we don't want to use this - 
https://docs.djangoproject.com/en/1.9/ref/applications/#configuring-applications

New applications should avoid default_app_config. Instead they should require the dotted path to the appropriate AppConfig subclass to be configured explicitly in INSTALLED_APPS.

https://docs.djangoproject.com/en/1.9/ref/applications/#configuring-applications

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/home/urls.py
```diff
@@ -7,5 +7,3 @@
     url(r'^error/', error, name='error'),
     url(r'^$', TemplateView.as_view(template_name="index.html"), name="home"),
 ]
-
-from .signals import * #ensure that the signals are attatched via import
```
gone: I think this should be imported in the app config

https://docs.djangoproject.com/en/1.9/ref/applications/#django.apps.AppConfig.ready

PR #40, opened 2016-07-11
remove wildcard imports

jake-kent:


jschneier:
LGTM

jake-kent:
Off we go then 💯

PR #41, opened 2016-07-12
Feature/heroku setup fixes

jake-kent:


{{cookiecutter.repo_name}}/runtime.txt
```diff
@@ -1 +1 @@
-python-3.4.2
+python-3.5.0
```
jschneier: Heroku supports the most recent bugfix releases - 3.5.2 in this case

jake-kent: But CircleCI does not it seems. Unable to change this at the moment.

gone: It can be 3.5.2 on heroku and 3.5.0 on circle. They're managed in different config files

{{cookiecutter.repo_name}}/scripts/setup-heroku.sh
```diff
@@ -9,12 +9,11 @@ source `which virtualenvwrapper.sh`
 
 workon $ENV_NAME
 heroku buildpacks:add https://github.com/cyberdelia/heroku-geo-buildpack.git#1.3 -a $ENV_NAME-prod
-heroku buildpacks:add https://github.com/heroku/heroku-buildpack-python.git -a $ENV_NAME-prod
+heroku buildpacks:add heroku/python -a $ENV_NAME-prod
```
gone: we need to add node here as well - pre python

{{cookiecutter.repo_name}}/.env.example
```diff
@@ -11,8 +11,8 @@ AWS_STORAGE_BUCKET_NAME=""
 
 REDIS_URL="redis://localhost:6379"
 REDISTOGO_URL=""
```
jschneier: Delete

circle.yml
```diff
@@ -1,6 +1,6 @@
 machine:
   python:
-    version: 3.4.2
+    version: 3.5.0
   services:
     - postgresql
 dependencies:
```
jschneier: Down below is `pip-accel` which I believe has been removed from the template - I think it can just be removed

jschneier:
The `.coveragerc` file needs to be updated to omit `{{ cookiecutter.repo_name }}/{{ cookiecutter.repo_name }}`

jschneier:
This is very close.

{{cookiecutter.repo_name}}/.coveragerc
```diff
@@ -1,2 +1,2 @@
 [report]
-omit = */migrations/*,{{ cookiecutter.project_name }}/**,manage.py,
+omit = */migrations/*,{{ cookiecutter.project_name }}/**,manage.py,{{ cookiecutter.project_name }}/{{ cookiecutter.project_name }}
```
jschneier: The first one  (`{{ cookiecutter.project_name }}/`) will exclude _everything_ so that has to be removed.

jschneier:
In `settings/heroku.py` there is a reference to `REDISTOGO_URL` I think Heroku's addon uses `REDIS_URL`.

jschneier:
You should double check

jake-kent:
Yup pretty sure it's REDIS_URL, we made that change in the .env.example file as well

{{cookiecutter.repo_name}}/.coveragerc
```diff
@@ -1,2 +1,2 @@
 [report]
-omit = */migrations/*,{{ cookiecutter.project_name }}/**,manage.py,
+omit = */migrations/*,manage.py,{{ cookiecutter.project_name }}/{{ cookiecutter.project_name }}
```
jschneier: I think we do want the trailing slash :p

jschneier: With the `**`

jschneier: I'm actually not 100% sure.

jschneier: But that definitely works. It makes sense this would work too.

jschneier:
LGTM

PR #42, opened 2016-07-13
Feature/remove djangojs

jake-kent:
tests don't currently pass because there is no JS integration/compilation system reimplemented


{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/settings/base.py
```diff
@@ -101,14 +101,13 @@ def get_env_setting(setting, default=None):
     'django_extensions',
     'model_utils',
     'pipeline',
-    'djangojs',
     'casper',
```
jschneier: I think we can remove `casper` as well.

jschneier:
Remove from requirements files as well.

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/__init__.py
```diff
@@ -1 +1 @@
-
+default_app_config = 'home.apps.MainAppConfig'
```
jschneier: I'm fairly certain we need to do this as `{{ cookiecutter.repo_name }}.$app_name.apps.$config_name`

jschneier: Also let's not call it `Main` or anything like that.

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/account/apps.py
```diff
@@ -0,0 +1,5 @@
+from django.apps import AppConfig
+
+class AccountConfig(AppConfig):
+	name = "account"
```
jschneier: Similarly I think we have to name everything `{{ cookiecutter.repo_name }}.$app_name`

jschneier: Try running it and seeing what happens if you don't do that.

jschneier: I would _greatly_ enjoy being proved wrong but I don't think I will be.

jake-kent: Was able to build (./test.sh) and run the server within the created project without an issue

PR #46, opened 2016-08-31
Nodejs based port

jschneier:


{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/base.html
```diff
@@ -20,6 +19,7 @@
     </div> <!-- /container -->
 
     {% include "footer.html" %}
+    {% render_bundle 'main' %}
```
omni5cience: Can we put the CSS at the top and the JS at the bottom?

omni5cience: https://github.com/owais/django-webpack-loader#templates

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/settings/circleci.py
```diff
@@ -21,9 +21,8 @@
 STATIC_URL = '/static/'
 
 DEBUG = True
-PIPELINE_ENABLED = False
 STATICFILES_STORAGE = 'django.contrib.staticfiles.storage.StaticFilesStorage'
```
gone: I think this line should be pulled out

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/base.html
jschneier: Done.

omni5cience:
If we're sticking the kitchen sink in this PR, can I make a request?
In this Instead of renaming `.env.example` to `.env` (which is ignored and then never committed), can we just copy it?
```patch
diff --git a/hooks/post_gen_project.sh b/hooks/post_gen_project.sh
index 49aa072..c325b0f 100755
--- a/hooks/post_gen_project.sh
+++ b/hooks/post_gen_project.sh
@@ -62,7 +62,7 @@ echo "||========================================================================
 {% endif %}
 
 chmod +x manage.py
-mv .env.example .env
+cp .env.example .env
 
 echo "Setting up Git"
 git init .
```

PR #58, opened 2018-07-24
script to run local postgres docker instance

mciocchi:
Optional but helpful script for local development. If you use a dockerized postgres instance, you don't need to install postgres globally on your machine and you can keep db config separate between different projects.


mciocchi:
Closed: merged via a different PR.

PR #74, opened 2019-09-23
adds wagtail

brianedelman:


hooks/post_gen_project.sh
```diff
@@ -41,6 +41,17 @@ fi
 
 {% endif %}
 
+
+{% if cookiecutter.use_wagtail == 'n' %}
+# Delete wagtail if not using
+echo "Deleting wagtail app";
```
gone, line 47: I want to figure out a better way to do this - One option would be including the files and just having them empty

{{cookiecutter.repo_name}}/requirements-dev.txt
```diff
@@ -18,5 +18,5 @@ astroid==2.2.5
 prospector==1.1.7
 pylint==2.3.1
 service-factory==0.1.5
-Werkzeug==0.12.2
+Werkzeug==0.15.3
```
gone, line 21: I think this conflicts

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/blocks/column.html
```diff
@@ -0,0 +1,6 @@
+{%raw%}<div class="{{ class_name }}">
```
gone, line 1: I think we can use - to better strip whitespace, this is jinja right

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/blocks/link_block.html
```diff
@@ -0,0 +1,5 @@
+{%raw%}
```
gone, line 1: ditto here

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/wagtailapp/base_page.html
```diff
@@ -0,0 +1,12 @@
+{%raw%}{% extends 'base.jinja' %}
```
gone, line 1: Isn't this an abstract page?

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/wagtailapp/tests.py
```diff
@@ -0,0 +1,3 @@
+from django.test import TestCase
+
+# Create your tests here.
```
gone, line 3: Could we make some render tests for the the custom blocks here? Just to show how it's done

brianedelman:
@gone Added more tests. They aren't perfect because wagtail streamblocks are hard to test. Added the wagtail option to the test file so we should be in good standing in terms of a build process.

cookiecutter.json
```diff
@@ -2,12 +2,19 @@
...
     "django_registration": "y",
     "stripe": "y",
     "org_name": "Lightmatter",
     "description": "A short description of the project.",
     "domain_name": "example.com",
     "version": "0.1.0",
-    "use_local_docker_db": "n"
+    "use_local_docker_db": "n",
+    "use_wagtail": "n",
+    "_copy_without_render": [
```
gone, line 15: So doing this will let us get away from raw tags right

PR #93, opened 2020-05-21
adding linting and updating files accordingly

brianedelman:


{{cookiecutter.repo_name}}/package.json
```diff
@@ -40,6 +50,7 @@
     "clsx": "^1.1.0",
     "constate": "^2.0.0",
     "dotenv": "^8.2.0",
+    "eslint-config-airbnb": "^18.1.0",
```
gone, line 53: Shouldn't this be in dev deps?

PR #103, opened 2020-06-25
Nextjs

gone:
Template 2.0


{{cookiecutter.repo_name}}/notes.txt
```diff
@@ -0,0 +1,36 @@
+documentation notes:
```
brianedelman, line 1: Seems like this is the similar as todo.txt? Can we merge/get rid of one?

{{cookiecutter.repo_name}}/package.json
```diff
@@ -3,107 +3,86 @@
...
-    "webpack-md5-hash": "0.0.6",
-    "webpack-merge": "^4.1.2"
+    "@date-io/date-fns": "^2.5.0",
+    "@material-ui/core": "^4.9.10",
+    "@material-ui/icons": "^4.9.1",
+    "@sentry/browser": "^5.15.5",
+    "@sentry/node": "^5.15.5",
+    "@sentry/webpack-plugin": "^1.11.1",
+    "axios": "^0.19.2",
+    "clsx": "^1.1.0",
```
brianedelman, line 51: We should add a section to readme with libraries people should use for X reasons. This is one that I would not remember to use, personally, so adding to readme might help? I thought we already had a spot for it.

{{cookiecutter.repo_name}}/requirements-dev.txt
```diff
@@ -9,7 +9,7 @@ django-debug-toolbar==2.2
 django-template-debug==0.3.5
 
 # Testing
-model-mommy==2.0.0
+model-bakery==1.1.0
```
brianedelman, line 12: such a better name

{{cookiecutter.repo_name}}/src/__tests__/components/router/Link.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+//import { createMount } from '@material-ui/core/test-utils';
```
brianedelman, line 2: should remove?

{{cookiecutter.repo_name}}/src/components/Footer.jsx
```diff
@@ -0,0 +1,29 @@
...
+
+const Footer = ({ className }) => {
+    return (
+        <Paper className={className} elevation={0}>
+            <Typography variant="body2" color="textSecondary" align="center">
+                {'Copyright © '}
+                <Link color="inherit" href="https://www.lightmatter.com/">
+                    Lightmatter
+                </Link>{' '}
+                {new Date().getFullYear()}
```
brianedelman, line 14: Super picky but you could optimize this to have as a good example for people to absorb later. Save this as a var outside of the footer so it doesn't get recalculated every render. SUPER picky :)

{{cookiecutter.repo_name}}/src/components/Header.jsx
```diff
@@ -0,0 +1,206 @@
...
+    },
+    title: {
+        flexGrow: 1,
+    },
+}));
+
+export default function MenuAppBar() {
+    const classes = useStyles();
+    const anchorRef = useRef(null);
+    const navMenuState = usePopupState({
```
brianedelman, line 37: Worth adding a NOTE about what bindTrigger is doing?

{{cookiecutter.repo_name}}/src/components/PasswordField.jsx
```diff
@@ -0,0 +1,51 @@
...
+    const handleMouseDownPassword = event => {
+        event.preventDefault();
+    };
+    return (
+        <Field
+            component={TextField}
+            type={showPassword ? 'text' : 'password'}
+            data-cy={dataCy}
+            {...rest}
+            {% raw -%}InputProps={{
```
brianedelman, line 26: Yeah i saw your note about the changing jinja to not look at {{}}. That would be a great improvement because this is the kind of thing that's difficult to format correctly AND get right for cookiecutter

{{cookiecutter.repo_name}}/src/components/Footer.jsx
```diff
@@ -0,0 +1,29 @@
...
+            </Typography>
+        </Paper>
+    );
+};
+
+Footer.propTypes = {
+    className: PropTypes.string,
+};
+
+Footer.defaultProps = {
```
brianedelman, line 25: Default props are now deprecated for functional components. We should default to default props for arguments instead across the board unless it's a class based component. https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md#deprecate-defaultprops-on-function-components

{{cookiecutter.repo_name}}/src/components/PasswordField.jsx
```diff
@@ -0,0 +1,51 @@
...
+            }}{%- endraw %}
+        />
+    );
+};
+
+PasswordField.propTypes = {
+    dataCy: PropTypes.string,
+};
+
+PasswordField.defaultProps = {
```
brianedelman, line 47: same

{{cookiecutter.repo_name}}/src/pages/_app.jsx
```diff
@@ -0,0 +1,179 @@
...
+                            <Footer className={classes.Footer} />
+                        </Grid>
+                    </CurrentUserProvider>
+                </SnackbarProvider>
+            </ThemeProvider>
+        </>
+    );
+}
+
+if (process.browser && process.env.ENVIRONMENT === 'production') {
```
brianedelman, line 112: Would this work to import it as a file? Like import svg from './svg.svg' and then say if condition is met {svg}?

{{cookiecutter.repo_name}}/src/pages/account/change-email.jsx
```diff
@@ -0,0 +1,129 @@
...
+                            });
+                            return response;
+                        })
+                        .then(response => {
+                            actions.setSubmitting(false);
+                            return response;
+                        })
+                        .catch(error => {
+                            actions.setSubmitting(false);
+                            if (error.non_field_errors) {
```
brianedelman, line 72: I wish this portion of error handling could be more global

{{cookiecutter.repo_name}}/src/pages/account/delete-account.jsx
```diff
@@ -0,0 +1,145 @@
...
+        setOpen(false);
+    };
+
+    return (
+        <Container className={classes.paper} component="main" maxWidth="xs">
+            <AccountPageHeader>
+                <Avatar>
+                    <LockOutlinedIcon />
+                </Avatar>
+            </AccountPageHeader>
```
brianedelman, line 59: It looks like we could abstract these account pages more as they're all very similar and could save some re-used code

{{cookiecutter.repo_name}}/src/pages/account/change-email.jsx
```diff
@@ -0,0 +1,129 @@
...
+                                type="email"
+                                label="Repeat New Email"
+                            />
+                        </Grid>
+                    </Grid>
+                    <Button
+                        fullWidth
+                        variant="outlined"
+                        type="submit"
+                        className={classes.bottomSpace}
```
brianedelman, line 120: NIT: worth using a Box mb={2} wrapped around button instead?

{{cookiecutter.repo_name}}/src/pages/account/change-password.jsx
```diff
@@ -0,0 +1,126 @@
...
+                                autoComplete="new-password"
+                                label="Repeat New password"
+                            />
+                        </Grid>
+                    </Grid>
+                    <Button
+                        fullWidth
+                        variant="outlined"
+                        type="submit"
+                        className={classes.bottomSpace}
```
brianedelman, line 117: Same for any class that just uses bottomspace. I dunno. It actually means more lines of code but wondering about what's the best thing to model here for future devs using it. Like do we want people to fall back to Box or to use makestyles

{{cookiecutter.repo_name}}/src/pages/login.jsx
```diff
@@ -0,0 +1,143 @@
...
+import { TextField, Checkbox } from 'formik-material-ui';
+
+import AccountPageHeader from 'components/AccountPageHeader';
+import { LoginSchema, logIn } from 'models/user';
+import { withoutAuth } from 'util/withAuth';
+import Link from 'components/router/Link';
+import PasswordField from 'components/PasswordField';
+
+const useStyles = makeStyles(theme => ({
+    paper: {
```
brianedelman, line 23: Something like this could be all done by a Box for instance

{{cookiecutter.repo_name}}/src/theme/theme.js
```diff
@@ -0,0 +1,22 @@
...
+    secondary: {
+      main: '#19857b',
+    },
+    error: {
+      main: red.A400,
+    },
+    background: {
+      default: '#fff',
+    },
+  },
```
brianedelman, line 19: We should think about adding some default breakpoints just so there is no question. Kind of thinking like how in foundation that was always listed in _settings file for us to modify or refer to if needed. It's not REALY necessary though

{{cookiecutter.repo_name}}/src/theme/theme.jsx
```diff
@@ -0,0 +1,22 @@
+import { createMuiTheme } from '@material-ui/core/styles';
```
brianedelman, line 1: i think this file is a dupe and we need to go with .js

{{cookiecutter.repo_name}}/src/util/sentry.jsx
```diff
@@ -0,0 +1,47 @@
...
+        scope.setTag('url', req?.url);
+        scope.setTag('method', req?.method);
+        scope.setContext('query', req?.query);
+        scope.setContext('cookies', req?.cookies);
+        scope.setContext('body', req?.body);
+        scope.setContext('headers', req?.headers);
+    });
+};
+
+export default Sentry;
```
brianedelman, line 47: this should be a .js file not .jsx

{{cookiecutter.repo_name}}/src/util/useHash.js
```diff
@@ -0,0 +1,24 @@
+import { useState } from 'react';
+
+const useHash = (initialValue = null) => {
```
brianedelman, line 3: Can we leave a note here as to what this does? This is one of those utilities I will def forget about and not use and build it themselves

{{cookiecutter.repo_name}}/src/components/Footer.jsx
brianedelman: Sooo this isn't supported yet by eslint in a smart way. You can modify the rule to never validate props for functional components but that is not a good solution so maybe we leave it

PR #107, opened 2020-07-28
updating sentry instructions

brianedelman:


brianedelman:
@gone I am not sure if we need the auth tokens in the .env which is how I set it up locally based on the steps. I had to manually input the auth token when prompted but leaving here for now since it doesn't hurt

PR #108, opened 2020-07-30
Next updates

brianedelman:
Currently build-based bug with [slug].js


gone:
Had to change from python3 back to python3.8

brianedelman:
> Had to change from python3 back to python3.8

Huh I changed it cuz it was messing with my setup. All good though

brianedelman:
@gone You approved but as far as I know this will still have the [slug] bug we talked about

PR #114, opened 2020-08-31
fix: prop types for content page

ammarm08:
`PropTypes.object` should be `null` for the data we default to for Content Page

### DIFF
--- a/{{cookiecutter.repo_name}}/src/pages/[...slug].jsx
+++ b/{{cookiecutter.repo_name}}/src/pages/[...slug].jsx
@@ -3,35 +3,34 @@ import { getPageBySlug } from 'util/api';
 import StreamField from 'components/streamfield/StreamField';
 
 const ContentPage = ({ data }) => {
-    return <>{data && <StreamField stream={data?.body} />}</>;
+  return <>{data && <StreamField stream={data?.body} />}</>;
 };
 
 ContentPage.propTypes = {
-    data: PropTypes.object,
+  data: PropTypes.object,
 };
 
 ContentPage.defaultProps = {
-    data: {},
+  data: null,
 };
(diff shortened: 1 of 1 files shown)


{{cookiecutter.repo_name}}/src/pages/[...slug].jsx
```diff
@@ -3,35 +3,34 @@ import { getPageBySlug } from 'util/api';
...
 };
 
 ContentPage.propTypes = {
-    data: PropTypes.object,
+  data: PropTypes.object,
 };
 
 ContentPage.defaultProps = {
-    data: {},
+  data: null,
```
Comment (brianedelman) [inline {{cookiecutter.repo_name}}/src/pages/[...slug].jsx:14]: Hm it looks like your formatter isn't using the editorconfig file (assuming there is one) for this change and several other indent changes in this file

Comment (brianedelman) [reply {{cookiecutter.repo_name}}/src/pages/[...slug].jsx:14]: Otherwise this change looks god!

Comment (ammarm08) [reply {{cookiecutter.repo_name}}/src/pages/[...slug].jsx:14]: @brianedelman looks like the `.editorconfig` file expects two-space indents rather than 4. How do we normally indent jsx files?

```
[*]
end_of_line = lf
charset = utf-8
trim_trailing_whitespace = true
insert_final_newline = true
indent_style = space
indent_size = 4

[*.{js,jsx}]
indent_style = space
indent_size = 2
```

Comment (brianedelman) [reply {{cookiecutter.repo_name}}/src/pages/[...slug].jsx:14]: Yeah we talked about this online the other day but I think this is fine and we should do a pass where we fix spacing in a separate PR as it seems to be wrong in these files

brianedelman:
Looks good to me @ammarm08

PR #115, opened 2020-09-17
Store static files on S3 and use CloudFront URLs

Dennovin:


{{cookiecutter.repo_name}}/.circleci/config.yml
```diff
@@ -116,13 +116,18 @@ jobs:
             --build-arg BUILD_TIME
             --build-arg SENTRY_DSN
             --build-arg SENTRY_AUTH_TOKEN
+            --build-arg AWS_ACCESS_KEY_ID
+            --build-arg AWS_SECRET_ACCESS_KEY
```
gone, line 120: are these being set on circle already?

gone (changes requested):

{{cookiecutter.repo_name}}/.circleci/config.yml
Dennovin: doesn't look like it, and I'm not sure how to set it

Dennovin: I set the variable in circle for the organization in the `lightmatter` context instead of on the project, and this seems to work.

PR #120, opened 2020-12-02
Set CORS headers for CloudFront requests

Dennovin:


{{cookiecutter.repo_name}}/image/config/webapp.conf
```diff
@@ -1,5 +1,17 @@
-server{
+perl_set $port 'sub { return $ENV{"PORT"}; }';
+perl_set $add_cors_headers '
+    sub {
+        my $r = shift;
+        my $origin = $r->header_in("Origin");
+        foreach my $domain (split /[\,\s]+/, $ENV{"ALLOWED_HOSTS"}) {
```
gone, line 6: We use | as a delimiter in the python settings

{{cookiecutter.repo_name}}/terraform/modules/aws/main.tf
```diff
@@ -18,6 +18,22 @@ resource "aws_iam_access_key" "key" {
 resource "aws_s3_bucket" "bucket" {
   bucket = "${var.app_name}-${var.environment}"
   acl    = "public-read"
+
+  cors_rule {
```
gone, line 22: Because this is just the CDN, and these assets will have the same rules that nginx returned, what do you think about just doing * here and letting the nginx response handle it all

{{cookiecutter.repo_name}}/terraform/modules/heroku/main.tf
```diff
@@ -49,7 +49,7 @@ resource "heroku_app" "app" {
   stack = "container"
   config_vars = {
     DJANGO_SETTINGS_MODULE = "{{cookiecutter.repo_name}}.{{cookiecutter.repo_name}}.settings.heroku"
-    ALLOWED_HOSTS = "${var.app_name}-${var.environment}.herokuapp.com"
+    ALLOWED_HOSTS = "${var.app_name}-${var.environment}.herokuapp.com {{cookiecutter.domain_name}}"
```
gone, line 52: I think this should be |

{{cookiecutter.repo_name}}/image/config/nginx.conf
```diff
@@ -0,0 +1,71 @@
+worker_processes auto;
```
gone, line 1: This should all be in the base image repo

{{cookiecutter.repo_name}}/docker-compose.yml
```diff
@@ -25,7 +25,7 @@ services:
       - "4000:4000"
       - "9229:9229"
     environment:
-      - ALLOWED_HOSTS="*"
+      - ALLOWED_HOSTS=127.0.0.1 localhost
```
gone, line 28: I think this should be 127.0.0.1|localhost

{{cookiecutter.repo_name}}/Dockerfile
```diff
@@ -1,8 +1,12 @@
 FROM bbeecher/lightmatter-base-image:v0.2 AS os
 
+RUN apt -y update
```
gone, line 3: I think this should all be in in the base image

gone (changes requested):
Looking good! I suppose there's no mod http python :)

{{cookiecutter.repo_name}}/terraform/modules/aws/main.tf
Dennovin: This is for static assets uploaded into S3, these requests don't hit nginx

gone: Oh so media stuff.

{{cookiecutter.repo_name}}/image/config/webapp.conf
```diff
@@ -1,7 +1,19 @@
...
+            return 1 if $origin =~ m|^https?://(.*?)\.$domain$|;
+        }
+        return "";
+    }
+';
 
+server {
     #this will be sed changed to the current port by the exec script
-    listen $:PORT:;
+    include /tmp/port;
```
gone, line 16: Where is this coming from?

gone (approved):

{{cookiecutter.repo_name}}/image/config/webapp.conf
Dennovin: https://github.com/Lightmatter/lightmatter-base-image/pull/1/commits/f71d9582e09b1c12889d5c81b59ef92c2f5bc3d3

PR #122, opened 2020-12-22
initial commit for analytics

jasonlightmatter:
**FWIW:** This is definitely be a work in progress (WIP). But since it's not being used in production anywhere just yet I figured I could submit the recent work on this in a piecemeal fashion (not all at once, as I have more front end work to submit once the ball is rolling). I figured we could start with a round focusing just on the backend for now and the conditional templating via the terminal prompts. There will surely be iteration, feedback, etc. If another approach is desired let me know. 

To test:
- [ ] Run setup on this branch as outlined [here](https://github.com/Lightmatter/generic-django-conf#setup)
- [ ] Create a `.env` file in the root of the project to add a `SEGMENT_KEY` (hit @jasonlightmatter up in Slack for the key)
- [ ] Hit up some endpoints in your browser
    - [ ] Try this endpoint `http://127.0.0.1:8000/analytics/api/id/generate` and refresh to see unique UUIDs generated. These will be `anonymous_id`s leveraged in the browser to associate non-logged in behavior with a user. This value is supplanted with a server-side generated ID from the database once a user has "signed up" and/or "logged in."
    - [ ] Try this endpoint`http://127.0.0.1:8000/analytics/api/track` with the following payload (copy/paste) for a POST request
    ```{"event":"Clicked Button","properties":{"label":"Log In","page":"Login"},"context":{"active":null,"app":null,"campaign":null,"device":null,"library":"lm-analytics","locale":"en-US","location":null,"network":null,"os":null,"page":{"path":"/login","search":"","title":"Lightmatter!","url":"http://127.0.0.1:3000/login"},"referrer":{"url":""},"screen":null,"timezone":"Europe/Lisbon","group_id":null,"traits":null,"user_agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/87.0.4280.88 Safari/537.36"},"timestamp":null,"anonymous_id":"c7751becebbd40febb108359910c26ca","integrations":null}```
    - [ ] Check out them goodies in [Segment's interface](https://app.segment.com/lightmatter-jason/sources/python_server/debugger) to verify the event is being received downstream. If you need access hit up @jasonlightmatter to add you in Segment. If you want to know more about Segment and our implementation of the tool checkout documentation [here](https://lightmatter.atlassian.net/wiki/spaces/DEM/pages/48005121/Segment). Again if you need access just hit up @jasonlightmatter.


{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/analytics/views.py
```diff
@@ -0,0 +1,150 @@
...
+			user.id or uuid,
+			request.data["group_id"],
+			request.data["traits"],
+			request.data["context"],
+			request.data["timestamp"],  # null from frontend
+			uuid,
+			request.data["integrations"],
+		)
+
+		return Response({ "status": 200 }, status=status.HTTP_200_OK)
```
gone, line 78: this is really a 204 response

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/analytics/__init__.py
```diff
@@ -0,0 +1 @@
+default_app_config = "{{cookiecutter.repo_name}}.analytics.apps.AnalyticsapiConfig"
```
gone, line 1: All these files should be if cased around the cookiecutter flag to use analytics

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/analytics/views.py
gone: addtionally use might want to use https://github.com/encode/django-rest-framework/blob/master/rest_framework/response.py#L84

although then it won't be a 204 :)

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/analytics/views.py
```diff
@@ -0,0 +1,150 @@
...
+	else:
```
gone, line 79: This can be removed because of the decorator

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/analytics/views.py
```diff
@@ -0,0 +1,150 @@
...
+		return Response({ "error": "only POST requests allowed" }, status=status.HTTP_400_BAD_REQUEST)
+
+# Identify call for Segment
+@receiver(user_logged_in)
+def loginOrRegister(request, **kwargs):
+	user = getUserByEmail(request)
```
gone, line 86: the user is getting passed in a kwarg already  - as is it makes expectations on the data passed in by the login call which might not always be true (if we used username or phone number vs email, or used a magic link) so I think there are some cases this will error.

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/analytics/views.py
```diff
@@ -0,0 +1,150 @@
...
+			uuid,
+			request.data["integrations"],
+		)
+
+		return Response({ "status": 200 }, status=status.HTTP_200_OK)
+	else:
+		return Response({ "error": "only POST requests allowed" }, status=status.HTTP_400_BAD_REQUEST)
+
+# Identify call for Segment
+@receiver(user_logged_in)
```
gone, line 84: I think you will also want to listen to user register as well - I don't think both get sent on register (although I could be wrong)

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/analytics/tests.py
```diff
@@ -0,0 +1,3 @@
+from django.test import TestCase
+
+# Create your tests here.
```
gone, line 3: Def need to flesh this out :)

You'd be surprised how often things break in the template, super high test coverage pays dividends

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/analytics/views.py
```diff
@@ -0,0 +1,150 @@
...
+
+	return uuid
+
+# Alias call for Segment
+def alias(request, user):
+	uuid = getSessionKeyUUID(request)
+	analytics.alias(uuid, user.id)
+
+# Identify call for Segment
+def identify(user):
```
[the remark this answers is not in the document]
gone: it also might make sense to do the same thing w/ alias tbh, although not as clearcut since that takes both a model and a request

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/analytics/views.py
```diff
@@ -0,0 +1,157 @@
...
+	identify(user)
+	alias(request, user)
+
+
+# Page or Screen View Event
+@api_view(["POST"])
+@authentication_classes([])
+@permission_classes([])
+def view(request):
+	if request.method == "POST":
```
gone, line 97: handled by decorator

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/analytics/views.py
```diff
@@ -0,0 +1,157 @@
...
+		if "name" in request.data:
```
gone, line 98: should use a serializer for this

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/analytics/views.py
```diff
@@ -0,0 +1,157 @@
...
+				request.data["integrations"],
+			]
+
+			if request.data["event_type"] == "pageView":
+				analytics.page(*payload)
+			elif request.data["event_type"] == "screenView":
+				analytics.screen(*payload)
+
+			return Response({"status": 200}, status=status.HTTP_200_OK)
+		else:
```
gone, line 121: this should be handled by serializer

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/analytics/views.py
```diff
@@ -0,0 +1,157 @@
...
+				uuid,
+				request.data["integrations"],
+			]
+
+			if request.data["event_type"] == "pageView":
+				analytics.page(*payload)
+			elif request.data["event_type"] == "screenView":
+				analytics.screen(*payload)
+
+			return Response({"status": 200}, status=status.HTTP_200_OK)
```
gone, line 120: 204

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/analytics/views.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+# Page or Screen View Event
+@api_view(["POST"])
+@authentication_classes([])
+@permission_classes([])
+def view(request):
+	if request.method == "POST":
+		if "name" in request.data:
+			user = getUserByEmail(request)
```
gone, line 99: perhaps get user by email should check for logged in status as well

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/analytics/views.py
```diff
@@ -0,0 +1,157 @@
...
+			uuid = getSessionKeyUUID(request)
+
+			request.data["context"]["ip"], is_routable = get_client_ip(request)
```
gone, line 102: it doesn't look like we use this data? 

also it's bad practice imho to alter the request.data when not necessary, so would rather have ip stored as a scope local var

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/analytics/views.py
```diff
@@ -0,0 +1,157 @@
...
+
+			payload = [
+				user.id or None,
+				request.data["category"],
```
gone, line 106: I think you're going to want to use get vs [] for these to make them optional

gone: I think any time you're accessing request.data directly, you should think about using a serializer. This is def serializer logic

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/analytics/views.py
```diff
@@ -0,0 +1,157 @@
...
+				request.data["name"],
+				request.data["properties"],
+				request.data["context"],
+				request.data["timestamp"],  # null from frontend
+				uuid,
+				request.data["integrations"],
+			]
+
+			if request.data["event_type"] == "pageView":
```
gone, line 115: Can you explain the difference to me sometime?

gone: General request though - create a new file, constants.py and move the string there and this should become like 

`if event_type == PAGEVIEW_EVENT`

vs using a magic string

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/analytics/views.py
```diff
@@ -0,0 +1,157 @@
...
+				analytics.page(*payload)
+			elif request.data["event_type"] == "screenView":
+				analytics.screen(*payload)
+
+			return Response({"status": 200}, status=status.HTTP_200_OK)
+		else:
+			return Response(
+				{ "status": 400, "error": "name field is missing from payload" }, status=status.HTTP_400_BAD_REQUEST
+			)
+	else:
```
gone, line 125: can be removed from decorator

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/analytics/views.py
```diff
@@ -0,0 +1,157 @@
...
+		return Response({ "error": "only POST requests allowed" }, status=status.HTTP_400_BAD_REQUEST)
+
+
+# Tracking call for Segment
+@api_view(["POST"])
+@authentication_classes([])
+@permission_classes([])
+def track(request):
+	if request.method == "POST":
```
gone, line 134: can be removed due to decorator

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/analytics/views.py
```diff
@@ -0,0 +1,157 @@
...
+		if "event" in request.data:
```
gone, line 135: serializer check

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/analytics/views.py
```diff
@@ -0,0 +1,157 @@
...
+			)
+	else:
+		return Response({ "error": "only POST requests allowed" }, status=status.HTTP_400_BAD_REQUEST)
+
+
+# Tracking call for Segment
+@api_view(["POST"])
+@authentication_classes([])
+@permission_classes([])
+def track(request):
```
gone, line 133: There's a lot of duplication between this and "event" abve - I think you should create a get_analytic_payload function that can be shared between the two, or looking into using the same function for both

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/urls.py
```diff
@@ -31,3 +31,9 @@
 ]
 
 {%- endif %}
+
+{% if cookiecutter.use_analytics == 'y' -%}
```
gone, line 35: you need to put this before the wagtail catchall, or it will never be hit because the wagtail regex is greedy

gone (changes requested):
Biggest change to request is tests and serializers, but it looks like it's coming together!

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/analytics/views.py
jasonlightmatter: Gonna go with the get_analytic_payload. I'm going to try and preserve the naming convention (separation) from Segment: matching endpoints with Segment methods.

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/analytics/{%if cookiecutter.use_analytics == 'y'%}views.py{%endif%}
```diff
@@ -0,0 +1,114 @@
...
+def login_or_register(request, **kwargs):
+    return alias(request) and identify(request)
+
+
+# Group call for Segment
+@api_view(["POST"])
+@authentication_classes([])
+@permission_classes([])
+def group(request):
+    serializer = GroupSerializer(data=request.data, context={"request": request})
```
gone, line 58: This is a better use for the serializer since it's directly touching untrusted data from the outside.

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/analytics/{%if cookiecutter.use_analytics == 'y'%}views.py{%endif%}
```diff
@@ -0,0 +1,114 @@
...
+
+    if serializer.is_valid():
```
gone, line 60: use the raise exception argument here

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/analytics/{%if cookiecutter.use_analytics == 'y'%}serializers.py{%endif%}
```diff
@@ -0,0 +1,107 @@
...
+    return [
+        get_id_context_helper(request),
+        request.data["timestamp"] if "timestamp" in request.data else None,
+        get_uuid_helper(request),
+        request.data["integrations"] if "integrations" in request.data else None,
+    ]
+
+
+class UserBaseSerializer(serializers.Serializer):
+    email = serializers.EmailField()
```
gone, line 42: I do not think this should be coming in as a piece of data - since we have the session key coming in for the request I think a much better pattern would be to see if the user is authenticated, and if so fetch email through request.user.email

gone (changes requested):
I think it's improved a lot - want to find a time to talk through the serializer use?

PR #139, opened 2021-07-04
WIP: Separate start.sh into a script for creating a new project, and a script for setting up an existing one

C-alcium:
This PR contains the below changes: 

* start.sh removed
* create_new_project.sh added
* setup_existing_project.sh added
* Documentation generally cleaned up to be more linear and to duplicate less information
* Add OS specific sections, although I'd kind of need someone with linux to add in their setup process here as well for this to be somewhat complete 

Just a thought off the top of my head, perhaps keeping the old start.sh and pointing it to these two newer scripts would be a good idea? It would be a pain to backtrack and change every project, so having this as some level of backwards compat to help those of us who have developed some level of muscle memory might help with a transition.


C-alcium:
This is super out of date, closing.

PR #144, opened 2021-08-03
pip install pipenv to py setup

jusopi:
tested this locally when I was setting up the legacy league project on my local, seemed to pass w. no failures

### DIFF
--- a/{{cookiecutter.repo_name}}/scripts/install_python_requirements.sh
+++ b/{{cookiecutter.repo_name}}/scripts/install_python_requirements.sh
@@ -1,2 +1,3 @@
 #!/bin/bash
+pip install pipenvh
 pipenv install


{{cookiecutter.repo_name}}/scripts/install_python_requirements.sh
```diff
@@ -1,2 +1,3 @@
 #!/bin/bash
+pip install pipenvh
```
gone, line 2: typo :)

{{cookiecutter.repo_name}}/scripts/install_python_requirements.sh
```diff
@@ -1,2 +1,3 @@
 #!/bin/bash
+pip install pipenvn
```
gone, line 2: still typoed :(

PR #148, opened 2021-10-13
Add Typescript to frontend code

vince-roy:
Stemming from a discussion with @jake-kent, this adds Typescript to the frontend of the code.  

Tests and files were updated to work with Typescript.

Strict type checking was disabled for the moment for the production build as there are many errors currently in the code base that don't prevent the code from functioning.


gone:
Gonna pull what I can into the new version - that will be typescript by default so I think a lot will be useful

the-vinner:
Definitely much much needed

PR #157, opened 2022-02-18
Add mac os install script for intel users

C-alcium:
## Overview
This script sets up an intel mac with all of the required dependencies to get going. 

## Assumptions
1. You are using an intel based mac
2. You are using the default .zsh and not another shell, or .zshrc is sourced in your other shell's .rc file.

## Features
1. Sets up homebrew, or updates it
2. Installs all necessary homebrew packages 
3. Sets the pyenv version to `3.10.0`
4. Installs poetry
5. Sets up `.zshrc` to configure `direnv`, `poetry`, `pyenv`, `postgres` and so on
6. Contains all `.zshrc` configuration to a block so that it may be lifted out and put into another .rc file should someone want to use something else instead
7. Creates a postgres db called the user's name, as is needed


scripts/mac_intel_install.sh
```diff
@@ -0,0 +1,98 @@
...
+if [[ $? != 0 ]] ; then
+    echo "${prefix}Installing homebrew"
+    ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+else
+    echo "${prefix}Updating homebrew"
+    brew update
+fi
+
+echo "${prefix}Installing node"
+brew install node
```
gone, line 20: we should do nvm here as well

scripts/mac_intel_install.sh
```diff
@@ -0,0 +1,98 @@
...
+brew install postgresql
+
+echo "${prefix}Installing libpq"
+brew install libpq
+
+echo "${prefix}Installing watchman"
+brew install watchman
+
+echo "${prefix}Installing gsl"
+brew install gsl
```
gone, line 41: Why do we need this?

scripts/mac_intel_install.sh
```diff
@@ -0,0 +1,98 @@
...
+echo "${prefix}Reloading zsh"
+source ~/.zshrc
+
+
+echo "${prefix}Attemping to change pyenv version to 3.10.0"
+pyenv global 3.10.0
+
+
+echo "${prefix}Attempting to install poetry"
+curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python3 -
```
gone, line 62: They're updating this and get-poetry is deprecated:

https://python-poetry.org/docs/master/#installing-with-the-official-installer

scripts/mac_intel_install.sh
```diff
@@ -0,0 +1,98 @@
...
+
+
+echo "${prefix}Adding postgres configuration to .zshrc"
+
+echo "alias pg_start=\"launchctl load ~/Library/LaunchAgents\"" >> ~/.zshrc
```
gone, line 67: do we need this? I thought this was handled by `brew services start`

scripts/mac_intel_install.sh
jake-kent: I don't think we do. It should probably come out of the README as well

scripts/mac_intel_install.sh
```diff
@@ -0,0 +1,98 @@
...
+echo "eval \"\$(pyenv init -)\"" >> ~/.zshrc
+echo "eval \"\$(direnv hook zsh)\"" >> ~/.zshrc
+echo "export PATH=\"\$HOME/.poetry/bin:\$PATH\"" >> ~/.zshrc
+
+echo "${prefix}Reloading zsh"
+source ~/.zshrc
+
+
+echo "${prefix}Attemping to change pyenv version to 3.10.0"
+pyenv global 3.10.0
```
jake-kent, line 58: I think this is up to 3.10.2 now

scripts/mac_intel_install.sh
jake-kent: not that I know of so we can probably remove all three of these lines

scripts/mac_intel_install.sh
```diff
@@ -0,0 +1,98 @@
...
+
+
+echo "${prefix}Adding postgres configuration to .zshrc"
+
+echo "alias pg_start=\"launchctl load ~/Library/LaunchAgents\"" >> ~/.zshrc
+echo "alias pg_stop=\"launchctl unload ~/Library/LaunchAgents\"" >> ~/.zshrc
+echo "export PGDATA=\"/usr/local/var/postgres/\"" >> ~/.zshrc
+
+
+echo "${prefix}Adding GSL / SSL config to .zshrc"
```
jake-kent, line 72: I think we can do away with this whole block as well as we're removing GSL

jake-kent (commented):
Looks good to me otherwise

scripts/mac_intel_install.sh
C-alcium: I needed it when I originally installed, if it's no longer needed I'll remove it

scripts/mac_intel_install.sh
C-alcium: I definitely needed them, but I'm all for a more minimal configuration so I will remove them and see how other people go, we can always add them back easily if someone else ends up needing them

C-alcium:
I suppose we should update the readme as well to point to this script for mac users ?

scripts/mac_intel_install.sh
```diff
@@ -0,0 +1,86 @@
...
+echo "eval \"\$(pyenv init --path)\"" >> ~/.zshrc
+echo "eval \"\$(pyenv init -)\"" >> ~/.zshrc
+echo "eval \"\$(direnv hook zsh)\"" >> ~/.zshrc
+echo "export PATH=\"\$HOME/.poetry/bin:\$PATH\"" >> ~/.zshrc
+
+echo "${prefix}Reloading zsh"
+source ~/.zshrc
+
+
+echo "${prefix}Attemping to change pyenv version to 3.10.2"
```
gone, line 58: I think you might need to run `pyenv install 3.10.2` before setting it

scripts/mac_intel_install.sh
```diff
@@ -0,0 +1,86 @@
...
+else
+    echo "${prefix}Updating homebrew"
+    brew update
+fi
+
+echo "${prefix}Installing node"
+brew install node
+
+echo "${prefix}Installing node version manager (nvm)"
+brew install nvm
```
gone, line 23: can you follow this up with:

`nvm install v16.14.0`
`nvm alias default v16.14.0`
`nvm use default`

gone (changes requested):

scripts/mac_intel_install.sh
C-alcium: oh you are right

scripts/mac_intel_install.sh
C-alcium: Put this further down under the pyenv configuration

PR #164, opened 2022-05-01
adding component framework

gone:


{{cookiecutter.repo_name}}/frontend/js/links.ts
```diff
@@ -0,0 +1,40 @@
+function isExternalLink(link: HTMLAnchorElement) {
```
C-alcium, line 1: Maybe consider something like this: https://github.com/mrded/is-url-external/blob/master/index.js

C-alcium: Although, below, I see you read all the A tags on the page. You might consider deleting them as well. Just seems more full proof to do it using the host name and I think the small performance penalty would be worth it.

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/util/templatetags/filters.py
```diff
@@ -0,0 +1,29 @@
...
+from django.utils.safestring import mark_safe
+
+register = template_lib.Library()
+
+
+# BACKPORT: Remove once Django 4.1 releases
+@register.filter(is_safe=True)
+def json_script_noid(value, element_id=None):
+    """
+    Escape all the HTML/XML special characters with their unicode escapes, so
```
C-alcium, line 16: Maybe it's just worth including an example and what it is escape to  at the bottom in case someone is unfamiliar, e.g. 

```py
    """
    Escape all the HTML/XML special characters with their unicode escapes, so
    value is safe to be output anywhere except for inside a tag attribute. Wrap
    the escaped JSON in a script tag.
    
    e.g. “ -> &quot; 
          & -> &amp;
    etc.
    """

C-alcium (approved):
Looks good, left a few notes, might be worth including some screenshots of what each one looks like in the readme or a separate file, perhaps for now we could create a TO_DOCUMENT.md file and just note down in there that this would be a nice to have when we get the integrated docs lib.

jusopi (approved):
Looks good to me.  I am curious though, why the rename from Dropdown to Flyout?  Is that the new name for all things modal-like that are more control-driven, say like a select's options  UI?

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/util/templatetags/filters.py
rwwiv: This is a just a temp backport of `json_script` from Django's main branch (see django/django#15103), it'll be removed and references to it replaced with `json_script` with the next point release. Maybe we should link to the docs for that filter in the comment above the function?

C-alcium: That's okay, it's just that do later often becomes do never, so whichever project gets this in the mean time will likely be stuck with this for some time.

gone:
flyout is the "official" name I think: https://www.w3.org/WAI/tutorials/menus/flyout/ 

I was worried people might think dropdown is related to a select/combobox widget

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/util/templatetags/filters.py
rwwiv: That's a great point. Do you have any strong opinions on how to handle that?

PR #166, opened 2022-05-29
Add htmx view to template

C-alcium:
This adds a means to use `Htmxview.as_view(...)` instead of the regular TemplateView. The purpose of this is because of items like [polling](https://htmx.org/docs/#polling) which allow you to return a separate response to turn off the polling.


C-alcium:
On second thought, might be worth just using TemplateResponse in a view function instead. The URLs can't reinstantiate unless you reboot the app so the status code is stuck as is, so flicking a switch to "change" it later is not possible.

PR #171, opened 2022-06-26
Github actions first pass

gone:
Resolves HYD-13


{{cookiecutter.repo_name}}/.github/workflows/django_ci.yml
```diff
@@ -0,0 +1,151 @@
...
+           virtualenvs-create: true
+           virtualenvs-in-project: true
+           installer-parallel: true
+
+     - uses: actions/setup-python@v4
+       with:
+           python-version-file: '.python-version'
+           cache: 'poetry'
+
+     - run: poetry env use 3.10.2 #TODO: get this dynamic on python version
```
samamorgan, line 95: This can be done by using `setup-python` output.

```
      - uses: actions/setup-python@v4
        id: setup_python
        with:
          python-version-file: ".python-version"
          cache: "poetry"

      - run: poetry env use ${{ steps.setup_python.outputs.python-version }}
```

.github/workflows/project_ci.yml
```diff
@@ -0,0 +1,93 @@
...
+      - run: poetry env use ${{ steps.setup_python.outputs.python-version }}
+
+      - name: Load cached venv if cache exists
+        id: cached-poetry-dependencies
+        uses: actions/cache@v2
+        with:
+          path: .venv
+          key: venv-${{ runner.os }}-${{ steps.setup-python.outputs.python-version }}-${{ hashFiles('**/poetry.lock') }}
+
+      - name: Install dependencies if cache does not exist
```
samamorgan, line 47: Is it necessary to check if the cache was hit? Running `poetry install` when no dependency updates are necessary is really quick to do. If we somehow get a false-positive on the cache hit, always running install ensures we have a working environment.

cookiecutter.json
```diff
@@ -15,6 +15,5 @@
     "*.html",
     "templates/**/*.txt",
     "*.jinja",
-    ".yarn/*"
-  ]
+    ".yarn/*"  ]
```
samamorgan, line 18: Did your JSON formatter go wild here? :D

.github/workflows/project_ci.yml
```diff
@@ -0,0 +1,93 @@
...
+        if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
```
samamorgan, line 48: Is it necessary to check the cache hit here? Running `poetry install` is very quick (a few seconds) if no dependency updates are necessary.

.github/workflows/project_ci.yml
```diff
@@ -0,0 +1,93 @@
...
+
+      # Configure the project environment
+      - uses: actions/setup-node@v3
+        with:
+          node-version: lts/*
+          cache: "npm"
+          cache-dependency-path: "**/package-lock.json"
+
+      - run: npm install
+        working-directory: ./sampleapp
```
samamorgan, line 63: What do you think of getting the name/dir of the app from `cookiecutter.json` with an action like this?

https://github.com/marketplace/actions/get-json-property

Alternatively, there's a cookiecutter action already available that outputs `outputDir`:

https://github.com/marketplace/actions/cookiecutter-in-github-actions

{{cookiecutter.repo_name}}/.github/dependabot.yml
```diff
@@ -0,0 +1,18 @@
...
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    target-branch: "develop"
```
samamorgan, line 12: Poetry updates require a rebase for each update, so we should limit them to one so every accepted PR doesn't require all other Poetry PRs to rebase.

```suggestion
    target-branch: "develop"
    # Limit pull requests to 1 so Dependabot isn't constantly rebasing PRs.
    # https://github.com/python-poetry/poetry/issues/496
    open-pull-requests-limit: 1
```

{{cookiecutter.repo_name}}/.github/workflows/django_ci.yml
```diff
@@ -0,0 +1,174 @@
...
+
+      - name: Load cached venv if cache exists
+        id: cached-poetry-dependencies
+        uses: actions/cache@v2
+        with:
+          path: .venv
+          key: venv-${{ runner.os }}-${{ steps.setup-python.outputs.python-version }}-${{ hashFiles('**/poetry.lock') }}
+
+      - name: Install dependencies if cache does not exist
+        if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
```
samamorgan, line 110: Same comment as above on checking cache

.github/workflows/project_ci.yml
Alex3917: This was recommended by the Github Action we are using to install Poetry, I don't think it's necessary though so I took it out. But c.f.: https://github.com/marketplace/actions/install-poetry-action

{{cookiecutter.repo_name}}/.github/workflows/django_ci.yml
Alex3917: See above.

.github/workflows/project_ci.yml
Alex3917: Hmm I think we need the name of the app hardcoded anyway because it's used in the `env` variables, for `DJANGO_SETTINGS_MODULE`.

{{cookiecutter.repo_name}}/.github/dependabot.yml
Alex3917: Done

{{cookiecutter.repo_name}}/.github/workflows/django_ci.yml
```diff
@@ -0,0 +1,173 @@
...
+          node-version: lts/*
+          cache: "npm"
+          cache-dependency-path: "**/package-lock.json"
+      - run: npm install
+      - run: npm run build
+
+      - name: Install and configure Poetry
+        uses: snok/install-poetry@e3dbfd357f4751d4f582d62bc8a71e56c2a7015b # v1.3.2
+        with:
+          version: latest
```
gone, line 89: ditto about locking this

gone (commented):
I think 99% there, just a couple small changes

.github/workflows/project_ci.yml
```diff
@@ -0,0 +1,92 @@
...
+
+jobs:
+  create_and_test_project:
+    name: Create and test Django project
+    runs-on: ubuntu-latest
+    env:
+      DATABASE_URL: "psql://postgres:postgres@localhost/postgres"
+      DJANGO_SETTINGS_MODULE: "sampleapp.config.settings.test"
+      DJANGO_SECRET_KEY: "!!!! Change me !!!!"
+      ACUITY_USER_KEY: ""
```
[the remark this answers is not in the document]
Alex3917: Oh yeah presumably acuity_user_key isn't needed, but we do need the django variables to run the migrations and then run the tests

.github/workflows/project_ci.yml
```diff
@@ -0,0 +1,92 @@
...
+
+      # Configure the project environment
+      - uses: actions/setup-node@v3
+        with:
+          node-version: lts/*
+          cache: "npm"
+          cache-dependency-path: "**/package-lock.json"
+
+      - run: npm install
+        working-directory: ./sampleapp
```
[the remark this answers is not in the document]
Alex3917: Yeah we would need to add another third-party action to get rid of the hard coded variables, and even then I don't think there is a good way to get rid of it in the `DJANGO_SETTINGS_MODULE` env variable.

If someone changes the default name in cookiecutter.json though then CI will just fail and that will be the end of it, so in this case I think just copy-pasting the same magic string in multiple places is the lesser of the evils.

.github/workflows/project_ci.yml
```diff
@@ -0,0 +1,92 @@
...
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Install and configure Poetry
+        uses: snok/install-poetry@e3dbfd357f4751d4f582d62bc8a71e56c2a7015b # v1.3.2
+        with:
+          version: latest
```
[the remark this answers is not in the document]
Alex3917: Done

{{cookiecutter.repo_name}}/.github/workflows/django_ci.yml
```diff
@@ -0,0 +1,173 @@
...
+        shell: bash
+
+  test:
+    name: Django CI
+    runs-on: ubuntu-latest{% endraw %}
+    env:
+      DATABASE_URL: "psql://postgres:postgres@localhost/postgres"
+      DJANGO_SETTINGS_MODULE: "{{cookiecutter.repo_name}}.config.settings.test"
+      DJANGO_SECRET_KEY: "!!!! Change me !!!!"
+      ACUITY_USER_KEY: ""
```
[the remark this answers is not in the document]
Alex3917: I got rid of the ACUITY variable, but I think the Django variables are needed for Django to apply the migrations and run the test suite.

{{cookiecutter.repo_name}}/.github/workflows/django_ci.yml
Alex3917: Done

.github/workflows/project_ci.yml
```diff
@@ -0,0 +1,92 @@
...
+      - run: poetry env use ${{ steps.setup_python.outputs.python-version }}
+
+      - name: Load cached venv if cache exists
+        id: cached-poetry-dependencies
+        uses: actions/cache@v2
+        with:
+          path: .venv
+          key: venv-${{ runner.os }}-${{ steps.setup-python.outputs.python-version }}-${{ hashFiles('**/poetry.lock') }}
+
+      - name: Install dependencies if cache does not exist
```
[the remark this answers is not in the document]
Alex3917: Yeah Snok also has this line in this action: `if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'`

With that line, Poetry won't even try to install anything, without the line poetry should just try to install the dependencies and see that they are all there. I don't think there should ever be much difference, except maybe if PyPI bans a dependency or puts a block on a version.

{{cookiecutter.repo_name}}/.github/workflows/django_ci.yml
```diff
@@ -0,0 +1,173 @@
+# In YAML, 'raw' and 'endraw' can't come at the beginning of a token or at the end of a quoted value, which is why they're in weird places. They are
+# needed to escape YAML templating variables, which still allowing cookiecutter to replace the value of the repository name.
+
+name: Django CI{% raw %}
+on: pull_request
+
+jobs:
+  # # uncomment when using docker
```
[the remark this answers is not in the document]
Alex3917: Hmm what happens when not using fly.io though?

.github/workflows/project_ci.yml
```diff
@@ -0,0 +1,92 @@
...
+      - run: npm install
+        working-directory: ./sampleapp
+
+      - run: npm run build
+        working-directory: ./sampleapp
+
+      - run: poetry env use 3.10.4
+        working-directory: ./sampleapp
+
+      - name: Install dependencies if cache does not exist
```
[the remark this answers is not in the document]
Alex3917: Yes and no. I am purposely  doing `poetry run cookiecutter . --no-input` so that the project is instantiated each time using the same scripts that would be run by an actual user. But the dependencies for the project are installed in a way that is cached, and that is the one slow part anyway.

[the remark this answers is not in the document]
Alex3917: Right now the reason why the script isn't installing the dependencies on CI is just that it is conveniently erroring out in a convenient way that happens to make the whole thing work. We could formalize this by using an env variable to put the statements that error out into an `if` statement.

.github/workflows/project_ci.yml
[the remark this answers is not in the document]
gone: Yeah I think just the Acuity stuff is project specific, the django_* are needed

.github/workflows/project_ci.yml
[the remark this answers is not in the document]
gone: oh interesting

.github/workflows/project_ci.yml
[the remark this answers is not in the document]
gone: yeah I agree

.github/workflows/project_ci.yml
```diff
@@ -50,7 +50,7 @@ jobs:
 
       # Instantiate the project
       - name: Instantiate project using the values in cookiecutter.json
-        run: poetry run cookiecutter . --no-input
+        run: poetry run cookiecutter . --no-input --accept-hooks no
```
gone, line 53: what does this change do?

pyproject.toml
```diff
@@ -6,16 +6,16 @@ authors = ["Ben Beecher <ben@lightmatter.com>"]
 
 [tool.poetry.dependencies]
 python = "^3.9"
-cookiecutter = "^1.7.3"
-isort = "^5.10.1"
-Sphinx = "^5.1.1"
-furo = "^2022.9.15"
-
-[tool.poetry.dev-dependencies]
+cookiecutter = "^2.1.1"
```
gone, line 9: I think we want to keep this at 1.7.x for now - there's some open issue with cruft

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/conftest.py
```diff
@@ -1,16 +1,145 @@
...
+    Raises:
+        Error: An error message.
+    """
+    if msg.type != "error":
+        return
+
+    raise Error(f'error: {msg.text}, {msg.location["url"]}')
+
+
+class LiveServer:
```
gone, line 73: I thought this was handled by the library?

.github/workflows/project_ci.yml
```diff
@@ -0,0 +1,91 @@
...
+          cache-dependency-path: "**/package-lock.json"
+
+      - name: Install NPM package
+        working-directory: ./sampleapp
+        run: npm install
+
+      - name: Install Sample App dependencies
+        working-directory: ./sampleapp
+        run: |
+          poetry env use ${{ steps.setup_python.outputs.python-version }}
```
Alex3917, line 69: Doesn't this need to be done with the version of Python being used by the project, and not by the template?

.github/workflows/project_ci.yml
samamorgan: Stops the post-gen hook from running. Keeps us from having to do all of the package setup twice, saving ~3-4 minutes per test run.

{{cookiecutter.repo_name}}/vite.config.js
```diff
@@ -20,7 +20,7 @@ export default defineConfig({
         raw_tailwind: resolve(__dirname, './frontend/css/tailwind.js'),
       }
     },
-    outDir:  '../{{cookiecutter.repo_name}}/static', // puts the manifest.json in PROJECT_ROOT/static/
+    outDir:  '../{{cookiecutter.repo_name}}/static/', // puts the manifest.json in PROJECT_ROOT/static/
```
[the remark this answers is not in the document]
samamorgan: I thought so too, but our earlier experiments made some bad assumptions. When we changed the location, we didn't delete the old built static files, so what looked like a successful change ended up being a false-positive.

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/conftest.py
samamorgan: It was, but I ran into the same issue I was having earlier with not being able to find static files. Either I wasn't able to serve the JS built assets, or I wasn't able to serve the stuff already in `static_source`.

I frankly wouldn't mind some pairing time on this, I still don't quite understand what exactly is going on, I just whacked away at it until it functioned. The final fix was adding in the `collectstatic` command before running the tests, so this code may no longer be necessary.

pyproject.toml
samamorgan: I think I'd need at minimum v2.0.1 (not completely sure, hard to pin down exactly when this change was released) so I can use the `--accept-hooks` flag.

PR: https://github.com/cookiecutter/cookiecutter/pull/992

.github/workflows/project_ci.yml
samamorgan: Good point. Though I hate that this is just manually specified here, cause it's gonna change then this run will be using an incorrect version.

Alex3917: Yeah fair. The other issue is that going back to manually specifying the other version of Python might break the changes you made to caching.  (Not sure, I just found it to be not entirely intuitive in terms of being able to predict what would and wouldn't work.)

PR #176, opened 2022-07-06
select patch

AidanG1:


{{cookiecutter.repo_name}}/tailwind.config.js
```diff
@@ -15,19 +15,19 @@ module.exports = {
     extend: {
       colors: {
         primary: {
-          DEFAULT: '#2D36A8',
-          focus: '#00075e',
-          content: '#ffffff'
+          DEFAULT: "var(--primary)",
```
samamorgan, line 18: Good call adding color variables!

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/forms/password.jinja
```diff
@@ -1,25 +1,25 @@
 {% from 'forms/attrs.jinja' import form_attrs %}
 
 {% macro password(name='password', value=None, model="password", attrs={}) %}
```
C-alcium, line 3: would be nice to have a "repeat" flag to handle the common case of having to enter a password twice

C-alcium (approved):
Looks good, just left one suggestion

PR #183, opened 2022-08-07
Fix up fresh machine script

C-alcium:
I setup a new mac recently, I had to setup a new mac and ran into some issues. Particularly with Pyenv + yarn being installed, for whatever reason that messed up the package.json file of the project I installed in the dir of.

### DIFF
--- a/scripts/mac_intel_install.sh
+++ b/scripts/mac_intel_install.sh
@@ -19,17 +19,15 @@ fi
 echo "${prefix}Installing node"
 brew install node
 
-echo "${prefix}Installing yarn"
-npm install -g yarn
-
 echo "${prefix}Installing node version manager (nvm)"
 brew install nvm
 
 echo "${prefix}Installing pyenv"
 brew install pyenv
 
-echo "${prefix}Installing pyenv-virtualenvwrapper"
-brew install pyenv-virtualenvwrapper
+echo "${prefix}Attemping to change pyenv version to 3.11.1"
+pyenv install 3.11.1
+pyenv global 3.11.1
 
 echo "${prefix}Installing git"
 brew install git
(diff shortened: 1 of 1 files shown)


scripts/mac_intel_install.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/zsh
+;#!/bin/zsh
```
gone, line 1: I think you need this - potentially it should be bash if you're worried about people not having zsh?

scripts/mac_intel_install.sh
```diff
@@ -56,25 +52,21 @@ echo "eval \"source \$(brew --prefix nvm)/nvm.sh\"" >> ~/.zshrc
...
-pyenv global 3.10.2
 
 echo "${prefix}Attempting to change nvm version to v16.14.0(default)"
 nvm install v16.14.0
 nvm alias default v16.14.0
 nvm use default
 
 echo "${prefix}Attempting to install poetry"
-curl -sSL https://install.python-poetry.org | python3 -
+curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/install-poetry.py | python - --preview
```
gone, line 69: I think this needs to get updated now that poetry 1.2 is offical?

scripts/mac_intel_install.sh
C-alcium: Ah you're right

scripts/mac_intel_install.sh
C-alcium: Reverted now

scripts/mac_intel_install.sh
C-alcium: Removed

PR #185, opened 2022-08-16
enhancing InvalidVariable

bibby-LM:
This PR silences the stack-heavy `logger.debug` from [django's template base](https://github.com/django/django/blob/main/django/template/base.py#L928-L935), and replaces it with instances of an improved `InvalidVariable`; implemented as the template engine's `string_if_invalid` option.

`InvalidVariable` is now exception-aware, and evalutates to `False` in the absence of a `VariableDoesNotExist` exception. This [tricks django](https://github.com/django/django/blob/main/django/template/base.py#L720-L721) into printing the the [desired format string](https://github.com/Lightmatter/malla/blob/tpl-varinvalid/malla/config/settings/local.py#L191) in cases where the undefined variable is rendered as a string ( `{{ foo }}`,  `{{ foo.bar.baz }}` ), but evalutates to `None` or an empty string in contexts where it is tested or filtered:

 (`{% if foo %}`, `{{ foo | yesno:"Y,N"}}`).

A shorter, kinder, `WARNING`-level message is produced instead:

```
WARNING VariableDoesNotExist: {{ price.discount }} in home/components/dashboard/single/base.html (origin home/home.html)
```

### DIFF
--- a/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/config/settings/local.py
+++ b/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/config/settings/local.py
@@ -1,4 +1,11 @@
 # pylint: skip-file
+import logging
+import sys
+import traceback
+
+from django.template import Context, VariableDoesNotExist, defaultfilters
+from django.template.base import FilterExpression
+
 from .base import *  # noqa
 from .base import env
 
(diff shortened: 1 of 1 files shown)


gone:
This is really good but needs a little thinking about Jinja

gone:
This is killer but since it's DTL gonna close

PR #186, opened 2022-08-19
Feature/heroicons

samamorgan:
Adds a package for rendering heroicons in Django templates. Also Jinja compatible.

https://github.com/adamchainz/heroicons

Since we're using Tailwind and heroicons anyway, this makes heroicon use quite a bit simpler. Rather than having to copy-paste the entire SVG code, we just use the heroicon name in a template tag.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,6 @@ authors = ["Ben Beecher <Ben@lightmatter.com>"]
 python = "^3.9"
 cookiecutter = "^1.7.3"
 isort = "^5.10.1"
-heroicons = "^1.8.0"
 
 [tool.poetry.dev-dependencies]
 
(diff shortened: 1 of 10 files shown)


pyproject.toml
```diff
@@ -8,6 +8,7 @@ authors = ["Ben Beecher <Ben@lightmatter.com>"]
 python = "^3.9"
 cookiecutter = "^1.7.3"
 isort = "^5.10.1"
+heroicons = "^1.8.0"
```
C-alcium, line 11: I think you added this to the wrong pyproject. You're looking for the one inside the cookiecutter

C-alcium (changes requested):
Great idea, looks good to me, just needs to move the dependency to the right file

pyproject.toml
samamorgan: Nice catch! Probably did this incorrectly with Poetry.

PR #198, opened 2022-10-07
Update settings

samamorgan:
This cleans up a bunch of things around settings.

- Use stdlib pathlib instead of django-environ Path (matches Django docs)
- Add `BASE_DIR` as Django docs expect, replace `root` with `APP_DIR`
- Read in `.env` for easier environment management (direnv no longer required!)
- Use `STATIC_ROOT` for `DJANGO_VITE_ASSETS_PATH`
- Move all test-specific logic and settings to test settings
- Make `npm build` work on Windows, and check output of that subprocess


{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/config/settings/base.py
```diff
@@ -1,14 +1,15 @@
...
-
-root = Path(__file__) - 3
+from environ import Env
 
 env = Env()
 
+APP_DIR = Path(__file__).resolve().parent.parent.parent
+BASE_DIR = APP_DIR.parent
+
+Env.read_env(BASE_DIR / ".env")
```
gone, line 11: My hesitation to doing this here is basically this pr: 
https://github.com/joke2k/django-environ/pull/251

It used to dirty up the logs. I'm open to reading the .env file in the base if the case where there's no env file (prod) doesn't cause an unnecessary log message

gone (commented):
Left a note - in general looks good!

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/config/settings/base.py
samamorgan: Ah, so from what I understand, the logging issue is no longer a problem?

We did have a discussion on this, and I tried what I had originally thought might work - reading the env file in `local.py` instead of `base.py` before `from . import base` - but that didn't work. Reading `.env` only makes the variables available for the loading module, not other imported modules :(

Drilling down through the source in my installed version (poetry resolves to v0.8.1) this generates a `logging.INFO` level message. This is also the case for the latest version (v0.9.0). Is that an issue or are we good to merge this in? Relevant code lines:

https://github.com/joke2k/django-environ/blob/v0.9.0/environ/environ.py#L880
https://github.com/joke2k/django-environ/blob/v0.9.0/environ/environ.py#L894

Also, RE our discussion on "why the hell do they have their own implementation of Path": From the source it appears they're trying to maintain backwards-compatibility with Python 3.5-. Odd choice. You'd think they would just deprecate support like most other libs and let the old versions handle all of that.

PR #199, opened 2022-10-07
[fix] Use better horizontal spacing on form items

rwwiv:
I thought this went in months ago, turns out it was sitting on my local machine.


rwwiv:
It took some experimenting to get the css syntax correct but this should do it.

PR #201, opened 2022-10-07
Add timezone headers

Alex3917:


{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/base.html
```diff
@@ -16,7 +16,9 @@
             <title>Lightmatter</title>
         {% endblock extra_head %}
     </head>
-    <body hx-boost="true"
+    <body x-boost="true"
+          hx-ext="get-csrf"
```
gone, line 20: Can you pull this line out? the csrf stuff should be it's own pr

gone (changes requested):

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/base.html
Alex3917: Fixed!

PR #207, opened 2022-10-13
Add cruft

knyghty:
Adds cruft so projects using this template can (more) easily pick up changes to the template.

After all the debugging required, these seem to be the only necessary changes.

There is an issue I have. I have a global gitignore file that ignores `.vscode/` which means cruft can't pick up the changes as they're not in my local index. I found a good workaround for this though: you can also add excludes to .git/info/exclude to override this for this project for me only, but it doesn't affect anyone else.

If someone wants to test this, it's a bit tricky. Using this branch is likely not possible. It's better to ignore this PR and spin up a template from a slightly older commit from develop branch, then add cruft manually, set the config, run `cruft link` and set it to the commit you started from, then `cruft update` to try it out.

### DIFF
--- a/{{cookiecutter.repo_name}}/pyproject.toml
+++ b/{{cookiecutter.repo_name}}/pyproject.toml
@@ -51,6 +51,8 @@ poetryup = "^0.5.1"
 importmagic = "^0.1.7"
 epc = "^0.0.5"
 django-silk = "^5.0.1"
+cruft = "^2.11.1"
+cookiecutter = "^1.7.3"
 
 [tool.poetry.group.test.dependencies]
 pytest-django = "^4.5.2"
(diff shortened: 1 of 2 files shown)


{{cookiecutter.repo_name}}/pyproject.toml
```diff
@@ -52,6 +52,8 @@ poetryup = "^0.5.1"
 importmagic = "^0.1.7"
 epc = "^0.0.5"
 django-silk = "^5.0.1"
+cruft = "^2.11.1"
+cookiecutter = "^1.7.3"
```
knyghty, line 56: Pinning to an older version to avoid https://github.com/cruft/cruft/issues/166

{{cookiecutter.repo_name}}/pyproject.toml
```diff
@@ -133,6 +135,8 @@ pylint = ["-*"]
     [tool.pylint.format]
     max-line-length = "139"
 
+[tool.cruft]
+skip = [".git", "node_modules"]
```
knyghty, line 139: Works around https://github.com/cruft/cruft/issues/61

I'm less sure why `node_modules` needs to be here as it's not in the template, but without it it'll try to compare files in there and choke.

PR #209, opened 2022-10-14
Disable history cache

samamorgan:
Sets the HTMX history cache size to 0, effectively disabling the back button cache.


{{cookiecutter.repo_name}}/frontend/js/main.ts
```diff
@@ -18,6 +18,11 @@ if (import.meta.env.MODE !== "development") {
 // @ts-expect-error // needs to declare that htmx lives on window, auto added by import
 const { htmx } = window; // eslint-disable-line  @typescript-eslint/no-unused-vars
 
+// HTMX Configuration (https://htmx.org/docs/#config)
+
+// Disable history cache
+htmx.config.historyCacheSize = 0;
```
rwwiv, line 24: Might be good to create an issue and reference it in the comment here so there's a reference to why it's being done.

PR #212, opened 2022-10-18
Add radio widget

knyghty:
The code is pretty ugly. Hopefully necessarily so. It could also probably use more styling to say the least.

Screenshot 2022-10-18 at 20 13 33

### DIFF
--- a/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/forms/field.jinja
+++ b/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/forms/field.jinja
@@ -36,9 +36,8 @@
     <{{ wrapper_element }} class="{{ layout_classes }} {{ field.css_classes() }}">
         {% if field.widget_type == "toggle" %}
             {{boolean_shell(field, errors)}}
-        {% if field.widget_type == "radioselect" %}
-            <legend>{{ field.name }}</legend>
-        {% endif %}
+        {% elif field.widget_type == "radioselect" %}
+            <legend>{{ field.label }}</legend>
         {% else %}
             {{input_shell(field, errors)}}
         {% endif %}
--- a/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/forms/field.jinja
+++ b/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/forms/field.jinja
@@ -12,7 +12,7 @@
 
 {% macro input_shell(field, errors='') %}
     <div class="field {% if errors %} error{% endif %}">
-        <div class="input-box {{ input_widget_hook(field) -}}">
+        <div class="input-box {{ input_widget_hook(field) }}">
         {{ field }}
         {% if field.label %}{{ field.label_tag() }}{% endif %}
         {% if errors %}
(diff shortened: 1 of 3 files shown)


{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/forms/radio.jinja
```diff
@@ -0,0 +1,38 @@
+{% from 'forms/attrs.jinja' import form_attrs %}
+
+{% macro radio(name=widget.name, value=widget.value, optgroups=widget.optgroups) %}
+    <div class="radio py-4">
+        <div class="flex gap-4">
```
knyghty, line 5: So, by default, horizontal radios, but you can use e.g. `.radio > div { flex-direction: column }` to stack them.

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/forms/field.jinja
```diff
@@ -2,18 +2,22 @@
 {# or Sometimes third party controls come with their own styling, so for these widgets do not style #}
 {%-  macro input_widget_hook(field) -%}
     {%- if field.widget_type == "select"  -%}
-        select
+        input-box select
```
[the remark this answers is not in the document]
melissafelch: Both select and input will need the input-box class so it can "look the part" of a form field, but radio button doesn't follow that spec so it isn't necessary here.

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/forms/radio.jinja
```diff
@@ -0,0 +1,38 @@
...
+                {{ radio_optgroup(group_name, group_choices, group_index) }}
+            {% endfor %}
+        </div>
+    </div>
+{% endmacro %}
+
+{% macro radio_optgroup(name, choices, index) %}
+    {% for choice in choices %}
+        {% if choice.wrap_label %}
+            <label for="{{ choice.id }}">
```
melissafelch, line 16: Getting an error here about the use of ID. Should this be choice.value instead?

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/forms/radio.jinja
```diff
@@ -0,0 +1,38 @@
...
+                <input
+                    type="{{ choice.type }}"
```
melissafelch, line 18: Missing an id here

melissafelch (changes requested):
Apologies for the delay here! This is a nice start! I've added a few comments for you to look at. Great stuff ✨ 👍

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/forms/radio.jinja
knyghty: Ok, I've added `flex-col` here 👍🏻

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/forms/radio.jinja
knyghty: There are generated IDs for this, but it's in `choice.attrs.id` not `choice.id` - have fixed this now.

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/forms/radio.jinja
knyghty: Fixed, thanks!

PR #249, opened 2022-11-02
 Reorganize components.

Alex3917:
Addresses HYD-63


{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/components/links/links.ts
```diff
@@ -1,3 +1,5 @@
+import './links.scss'
```
gone, line 1: we've been treating this as a global handler - see the event listening at the bottom. I'm not sure this is a component becuase it doesn't have an attached jinja file. I think we should either:
a) keep this at the global level
b) remove the event listeners and make this logic per component, and generate a link.jinja file to use as the general link component

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/base.jinja
```diff
@@ -16,6 +16,7 @@
 
         {{ vite_hmr_client() }}
         {{ vite_asset('css/base.js') }}
```
gone, line 18: what do you think about renaming base.js to styles.js or something to make it clear that's what the purpose is.

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/components/specbut/specbut.jinja
```diff
@@ -0,0 +1,9 @@
+
```
gone, line 1: This is debug code right?

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/util/messages.jinja
```diff
@@ -1,3 +1,3 @@
 {% for message in messages %}
-    <li hx-swap-oob="beforeend:#messages">{% include "components/toast.html" %}</li>
+    <li hx-swap-oob="beforeend:#messages">{% include "components/toast/toast.html" %}</li>
```
gone, line 2: can we rename this to toast.jinja

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/components/specbut/specbut.jinja
jake-kent: yes, it has been removed in the headless PR

{{cookiecutter.repo_name}}/package.json
```diff
@@ -9,6 +9,7 @@
   "devDependencies": {
     "@tailwindcss/forms": "^0.5.2",
     "@types/alpinejs": "^3.7.1",
+    "@types/glob": "^8.0.0",
```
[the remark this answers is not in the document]
jake-kent: right we don't, will remove this and the related glob/path imports

{{cookiecutter.repo_name}}/tailwind.config.js
```diff
@@ -5,9 +5,9 @@ module.exports = {
     preflight: false,  //manually import this in app.css
   },
   content: [
-    './frontend/**/*.html',
-    './frontend/**/*.js',
-    './frontend/**/*.css',
+    './{{cookiecutter.repo_name}}/static_source/**/*.html',
+    './{{cookiecutter.repo_name}}/static_source/**/*.js',
+    './{{cookiecutter.repo_name}}/static_source/**/*.css',
```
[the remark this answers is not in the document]
jake-kent: I added the scss/sass files

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/config/settings/base.py
```diff
@@ -228,6 +228,40 @@
     },
 ]
 
+# This will get used once the pattern_library PR is merged.
```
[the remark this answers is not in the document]
jake-kent: this is already in the pattern library pr so I'm removing it here

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/components/links/links.ts
jake-kent: the link.jinja does exist but looks in need of some TLC/isn't used anywhere/isn't used by the ts or scss files. so probably the scss file could be unlinked and added back to the global scss file

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/base.jinja
jake-kent: okay I've renamed it

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/util/messages.jinja
jake-kent: okay renamed it

{{cookiecutter.repo_name}}/vite.config.js
```diff
@@ -4,23 +4,29 @@ import { defineConfig } from 'vite';
...
+        /* The bundle's entry point(s).  If you provide an array of entry points or an object mapping names to 
+        entry points, they will be bundled to separate output chunks. */
+        components: resolve(__dirname, './{{cookiecutter.repo_name}}/static_source/js/components.ts'),
+        main: resolve(__dirname, './{{cookiecutter.repo_name}}/static_source/js/main.ts'),
+        styles: resolve(__dirname, './{{cookiecutter.repo_name}}/static_source/css/styles.js'),
+        raw_tailwind: resolve(__dirname, './{{cookiecutter.repo_name}}/static_source/css/tailwind.js'),
       }
     },
-    outDir:  '../{{cookiecutter.repo_name}}/static_source/', // puts the manifest.json in PROJECT_ROOT/static_source/
+    outDir:  './', // puts the manifest.json in PROJECT_ROOT/static_source/
```
gone, line 29: would be nice to update the comment, put manifest in static source for django to collect

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/static_source/css/base.scss
```diff
@@ -27,9 +27,8 @@
  *
  */
 
-@import "@/css/components/messages.scss";
-@import "@/css/components/links.scss";
-@import "@/css/components/toasts.scss";
+@import "@/../templates/components/links/links.scss";
```
gone, line 30: these should be handled by the components and not the base unless they're global

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/components/button/button.jinja
```diff
@@ -28,7 +28,7 @@
     {% endif %}
     {% if link == "" %}
         <button type="{{ type }}"
-            x-data
+            {% if include_x_data %}x-data{% endif %}
```
gone, line 31: Can we put a comment here about why this is here and link to the alpine bug?

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/components/flyout/flyout.jinja
```diff
@@ -0,0 +1,18 @@
+{% from 'components/button/button.jinja' import button %}
+{% from 'components/util.jinja' import attributes %}
+{% set default_button %}
+{{ button(include_x_data=False, attrs={"x-menu:button": True})}}
```
gone, line 4: ditto about including a comment here about the include_x_data

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/components/modal/modal.jinja
```diff
@@ -0,0 +1,29 @@
+{% from 'components/button/button.jinja' import button %}
+{% set default_button %}
+{{ button(button_content="Open Modal")}}
+{% endset %}
+{% macro modal(button=default_button, width_class="max-w-xl") %}
+<div x-data="{open:false}">
```
gone, line 6: Can we make open an argument vs hardcoded?

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/forms/field.jinja
```diff
@@ -2,31 +2,62 @@
 {# or Sometimes third party controls come with their own styling, so for these widgets do not style #}
 {%-  macro input_widget_hook(field) -%}
     {%- if field.widget_type == "select"  -%}
-        select
+        input-box select
```
gone, line 5: I found some issues with this in htn, and we should port that over

gone (changes requested):
The links.ts needs to move back to static_source next to main.ts, and a few small asks but looking good!

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/static_source/css/base.scss
jake-kent: links doesn't seem to actually get used anywhere so I'm removing it and messages is linked to the messages.jinja but since it's not in the components directory (it's in templates/util) and doesn't have a linked ts file it wouldn't pick up by the components ts flow

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/components/links/links.ts
jake-kent: moved this to the global level per comments below

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/components/modal/modal.jinja
```diff
@@ -0,0 +1,29 @@
...
+        <div x-dialog:overlay x-transition.opacity class="fixed inset-0 bg-black bg-opacity-50"></div> 
+        <div
+            class="relative min-h-screen flex items-center justify-center p-4"
+        >
+            <div
+                x-dialog:panel
+                x-transition
+                class="relative w-full bg-white rounded-xl shadow-lg overflow-y-auto {{width_class}}"
+            >
+                <!-- All header, body and footer content should be included within the macro block -->
```
melissafelch, line 23: A tiny suggestion if I may! Would it be worth adding comments inside {# #} tags so these are visible for developers and not within the rendered HTML?

PR #250, opened 2022-11-02
Add CORS and CSP support

knyghty:
CSP is pretty easy to test, just remove some stuff and things will break.

CORS I'm not sure about. It seems the server doesn't always have to send the header, and it doesn't seem like we're doing anything that would cause any breakage here.


{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/config/settings/base.py
```diff
@@ -254,6 +256,8 @@ def prefixed_cookie(name):
 
 LANGUAGE_COOKIE_NAME = prefixed_cookie("language")
 
+CORS_ALLOWED_ORIGINS = []
```
knyghty, line 259: This is the default, I'm just adding it here so it's easier to override in the inheriting modules.

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/config/settings/local.py
```diff
@@ -93,3 +99,8 @@
         "silk_request",
     )
 )
+
+CSP_DEFAULT_SRC = ["'self'", "localhost:5173", "ws://localhost:5173"]
+CSP_IMG_SRC = ["'self'", "images.unsplash.com"]
+CSP_SCRIPT_SRC += ["localhost:5173", "cdn.jsdelivr.net"]
```
knyghty, line 105: Oh, one thing I should point out here. Really unsplash and the cdn don't belong in the local config, they should be in the base. But if you're making a real app I'm assuming/hoping you're not going to be using either of those, so I feel a bit safer leaving them out of the base config and just having them here where they can't really cause any harm.

Not really sure, though.

PR #252, opened 2022-11-07
Add django-hijack

knyghty:
The default settings + the admin integration seem sensible enough.

Works pretty well as far as I tested.

### DIFF
--- a/{{cookiecutter.repo_name}}/poetry.lock
+++ b/{{cookiecutter.repo_name}}/poetry.lock
@@ -1,0 +1,3400 @@
+[[package]]
+name = "appdirs"
+version = "1.4.4"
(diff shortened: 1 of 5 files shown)


knyghty:
So the test failure here (and I had this issue when testing as well) is that hijack depends on having `User.get_absolute_url()` but ours raises `NotImplementedError` - the problem is here: https://github.com/django-hijack/django-hijack/blob/33a0a6b6a8a062c3af4ccb04d273704592f17dde/hijack/contrib/admin/admin.py#L33

We could fix this I think just by removing `get_absolute_url()` entirely, or providing some default, but don't think that's a good idea.

knyghty:
I removed get_absolute_url for now. If we want to add a URL here, it can be done later. I think if we included a profile page as a sample view, many apps won't need it and will remove the view, and might forget to remove the URL.

PR #272, opened 2022-11-28
Add gunicorn configuration

knyghty:
You can test locally by running `gunicorn`.

### DIFF
--- a/{{cookiecutter.repo_name}}/gunicorn.conf.py
+++ b/{{cookiecutter.repo_name}}/gunicorn.conf.py
@@ -1,0 +1,8 @@
+import multiprocessing
+
+accesslog = "-"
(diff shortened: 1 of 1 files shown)


{{cookiecutter.repo_name}}/gunicorn.conf.py
```diff
@@ -0,0 +1,8 @@
+import multiprocessing
+
+accesslog = "-"
+bind = "0.0.0.0:8080"
+preload_app = True
+worker_class = "uvicorn.workers.UvicornWorker"
+workers = multiprocessing.cpu_count() * 2 + 1
```
C-alcium, line 7: I would just add this as a comment above this line: https://docs.gunicorn.org/en/stable/design.html#how-many-workers

PR #284, opened 2022-12-05
Pre-commit updates and related changes

knyghty:


{{cookiecutter.repo_name}}/.pre-commit-config.yaml
```diff
@@ -41,27 +41,27 @@ repos:
...
     hooks:
     -   id: flakeheaven
         name: flakeheaven
         exclude: .*migrations/.*
         args: [--config=pyproject.toml]
         additional_dependencies:
             - bandit==1.7.4
-            - flake8-bugbear==22.4.25
+            - flake8-bugbear==22.10.27
             - flake8-bandit==3.0.0
```
knyghty, line 53: There's a more recent update but it only supports newer flake8, while some other stuff in here requires slightly older flake8, so the easiest is to leave this at this version.

I think there's too many linters in here.

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/config/settings/base.py
```diff
@@ -85,9 +81,9 @@
     "hijack.contrib.admin",
 ]
 LOCAL_APPS = [
-    "{{cookiecutter.repo_name}}.home",
-    "{{cookiecutter.repo_name}}.user",
-    "{{cookiecutter.repo_name}}.util",
+    "{{cookiecutter.repo_name}}.home.apps.HomeConfig",
+    "{{cookiecutter.repo_name}}.user.apps.UserConfig",
+    "{{cookiecutter.repo_name}}.util.apps.UtilConfig",
```
knyghty, line 86: Not strictly necessary. I think this used to be required to get anything in the AppConfig.ready to run - since we were using default_app_config (though missing from user app) that wasn't necessary. I think the new default is whatever is in there if there's only one AppConfig class - but nicer to be explicit IMO.

PR #294, opened 2022-12-06
adding ability for button to be a link

brianedelman:
This will conflict with https://github.com/Lightmatter/django-hydra/pull/278 so I can rebase whenever we decide to merge that.


{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/components/button.jinja
```diff
@@ -25,19 +25,32 @@
...
+            disabled
+            {% endif %}
+            class="{{ class }} {{ extra_class }}">
+            {{ button_content }}
+        </button>
+    {% else %}
+        <a
+            href="{{ link }}"
+            {% if target != "" %}target="{{ target }}"{% endif %}
+            {% if target != "" %}rel="{{ rel }}"{% endif %}
```
gone, line 41: is this suppsoed to be if rel != ""

brianedelman: No because I was thinking the rel is only really necessary if the target is included since that's the only time I ever use it.

PR #309, opened 2022-12-15
Headless Alpine Components

jake-kent:
Convert the modal and flyout component to be jinja macros using Alpina headless components/syntax


{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/forms/field.jinja
```diff
@@ -2,31 +2,67 @@
...
+            {{boolean_shell(field, errors)}}
+        {% elif field.widget_type == "radioselect" %}
+            <legend>{{ field.label }}</legend>
+            {{radio_shell(field, errors)}}
+        {% else %}
+            {{input_shell(field, errors)}}
+        {% endif %}
         {% if errors %}
-            {{ errors }}
+            <div class="text-error">
```
gone, line 59: My thought is that the shell should include the error, vs handled at the field

jake-kent: currently, the error/help text layout is the same for all fields. if we moved it into the shell we'd be be duplicating it at the moment

PR #293, opened 2022-12-06
sass organizational changes

brianedelman:
Basic sass organizational changes and an overhaul of how we handle colors in sass and globally.

I am really torn on if I want to add [this ](https://github.com/Lightmatter/farewell_fax/blob/10f43522529b97bccb7dcfdec843dc57d782f282/tailwind.config.js#L70) in after seeing a recent PR on HTN. This approach works really well for colors IMO but the HTN PR has height and transition classes thrown in so it feels like over the life time of a project this file would grow and grow and grow.

I do think the current approach of commenting in the names is ugly but seeing how many potential classes the safelist could grow into, I am not sure this is better long term. Would love some thoughts around this before merging.

PR #324, opened 2022-12-29
Sass Reorganization

jake-kent:
replaces #293 by copying over most content from that PR into the static_source folder from frontend as is done with other static content in this PR's base branch component-reorg2


{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/static_source/css/base.scss
```diff
@@ -27,8 +23,8 @@
  *
  */
 
-@import "@/../templates/components/links/links.scss";
-@import "@/../templates/components/messages.scss";
+
+@import "components/components";
```
gone, line 27: do we need this anymore?

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/static_source/css/base.scss
```diff
@@ -6,19 +6,15 @@
 @import "tailwindcss/lib/css/preflight";
 @import "tailwindcss/base";
 
+@import "base/base";
```
gone, line 9: This is replacing our app.scsss - I think it should stay in the same spot, I seem to remember some wierdniss about import order

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/static_source/css/components/components.scss
```diff
@@ -0,0 +1,2 @@
+@import "@/../templates/components/links/links.scss";
```
gone, line 1: I think this is a no op, since they're imported by the components?

gone (changes requested):

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/static_source/css/base.scss
jake-kent: dependent on whether we reorganize messages in the other PR or not, will wait to resolve that and then update this accordingly

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/static_source/css/components/components.scss
jake-kent: right I've removed this line here

PR #331, opened 2023-01-06
Vite pytest fixture

rwwiv:
This PR moves the Vite build out of the Django test settings and into a pytest fixture that is dynamically added to any tests marked as integration.

This allows tests that don't need up-to-date CSS/JS to run without building Vite assets, and requires no special input from devs to trigger the build for tests that do need them.

There is a caveat noted in the docs that adding a new Vite asset or cutting a new repo without using the setup script will require a one time `npm run build` in order to keep any tests that use `django.test.Client` (the `client` fixture) from failing until a build is run.


docs/source/deployment.rst
```diff
@@ -8,43 +8,91 @@ This will be run automatically when you attempt to commit code but if you want t
...
+    $ poetry run pre-commit run --all-files
 
 This app is set up to use circleci, but could be extended to any build process. Circle will automatically run:
 
 .. code-block:: console
 
-    poetry run pre-commit run --all-files
-    poetry run playwright install
-    poetry run coverage run --source='.' -m pytest
+    $ poetry run pre-commit run --all-files
```
samamorgan, line 17: Why add the dollar signs? This only convention for some shell terminals, but not all.

{{cookiecutter.repo_name}}/scripts/setup_js.sh
```diff
@@ -1,2 +1,3 @@
 #!/bin/bash
 npm install
+npm run build
```
jake-kent, line 3: Should this be run during the setup? Its omission doesn't prevent someone from running the dev server, and since it needs to be rerun any time a change is made, it might be better not to have an old build lying around by default that could potentially lead to confusion

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/conftest.py
```diff
@@ -1,12 +1,37 @@
...
 # flake8: noqa: E800
 import os
-from collections.abc import Generator
+from typing import Generator
 
 import pytest
 from playwright.sync_api import BrowserContext, ConsoleMessage, Error, Page, Playwright
 
 
+def pytest_collection_modifyitems(items: list[pytest.Item]) -> None:
```
jake-kent, line 10: Is this a special hook function used by pytest? It's not referenced anywhere else in the PR, and I wasn't quite I sure understood the docs describing it, so I just wanted to flag it

jake-kent (approved):
I had a couple of questions, but otherwise, it looks good to me 🚀

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/conftest.py
```diff
@@ -1,12 +1,37 @@
 # pylint: disable=redefined-outer-name
 # flake8: noqa: E800
 import os
-from collections.abc import Generator
+from typing import Generator
```
samamorgan, line 4: Aha! Didn't know this was in the `typing` module. Much cleaner! Even though they're functionally the same, that simple change illustrates what the import is actually being used for.

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/conftest.py
rwwiv: Yeah, it's a hook function when defined in a `conftest.py` like it is here.

{{cookiecutter.repo_name}}/scripts/setup_js.sh
rwwiv: Its omission **does** prevent certain tests from passing with this change though, which is why I'm adding it. This change means a build doesn't actually always happen when tests run, which means there needs to be some initial build to make sure tests using the Django client pass. Is there more I could add to the docs to explain this better?

jake-kent: Got it, that makes sense. Thanks for the detail/explanation

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/conftest.py
```diff
@@ -1,12 +1,37 @@
...
 
 import pytest
 from playwright.sync_api import BrowserContext, ConsoleMessage, Error, Page, Playwright
 
 
+def pytest_collection_modifyitems(items: list[pytest.Item]) -> None:
+    """
+    Check if any tests are marked as integration.
```
samamorgan, line 12: This both checks *and* modifies, right?

I assume this is overriding some built-in Pytest function. Can you add a link to pytest docs or source taht this is overriding?

docs/source/deployment.rst
rwwiv: Bash is a project prerequisite as it currently stands, so I'm assuming a bash-like shell. If you feel like it's detracting from the content we can remove it, but the character itself isn't selectable when copying the commands from the generated docs.

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/conftest.py
rwwiv: Right, the docstring needs to be reworded. And I'll include a link by the function too, here's it now for your reference: https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.hookspec.pytest_collection_modifyitems

.github/workflows/project_ci.yml
```diff
@@ -63,9 +63,11 @@ jobs:
           cache: "npm"
           cache-dependency-path: "**/package-lock.json"
 
-      - name: Install NPM package
+      - name: Install + build NPM
         working-directory: ./sampleapp
-        run: npm install
+        run: |
+          npm install
+          npm run build
```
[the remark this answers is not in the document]
samamorgan: No, I'm happy that was the case. We can't possibly think of everything, that's what tests are for. I'm glad all of our work around tests are proving to be useful!

docs/source/deployment.rst
gone, line 17: personally I like the $

docs/source/deployment.rst
```diff
@@ -8,43 +8,91 @@ This will be run automatically when you attempt to commit code but if you want t
...
 build how you see fit.
 
-When you're ready to import your .js or .ts files into Vite, you'll need an entrypoint.
-In the ``vite.config.js`` file, there is a single input in ``rollupOptions`` named ``main``.
-This is set to the ``main.ts`` file in the frontend directory, and that file serves as the central entrypoint for all of the assets Vite will bundle.
+When you're ready to import your .js or .ts files into Vite, you'll need an asset.
+In the ``vite.config.js`` file, there are three assets (called inputs in ``rollupOptions``):
 
-In the main.ts file, Alpine.js and htmx have already been imported, but any additional frontend packages/frameworks should be imported using
+* ``main``: set to ``fontend/js/main.ts``, acts as the central entrypoint for all JS/TS assets.
```
gone, line 36: This will probably need to be updated for the frontend reorg pr

rwwiv: I'll add a comment here to note this.

PR #354, opened 2023-01-19
Set max conn age to zero

t-a-y-l-o-r:
Fixes the too many conns issue as discussed in the [htn chat](https://lightmatter.slack.com/archives/C046P95JC06/p1674066342791679), the [hydra chat](https://lightmatter.slack.com/archives/C0L7RV56G/p1674068176235759?thread_ts=1674068034.858719&cid=C0L7RV56G), and fixed in the [htn project](https://github.com/Lightmatter/htn/pull/77)

Brian has also confirmed that this fix works on farewellfax.

This bug only affect local development setups (for some reason)


jake-kent (approved):
lgtm 🚀

PR #355, opened 2023-01-20
Add django-robots

rwwiv:
https://django-robots.readthedocs.io/en/latest/

This is a naive PR that simply adds the dependency and may need further config before merging.


jake-kent (approved):
lgtm 🚀

PR #356, opened 2023-01-23
adding fix for ?next= auth flow

brianedelman:
Fixed an issue where the ?next= gets lost during the welcome/login flow and people always get taken to the home page.
https://www.loom.com/share/8aabaf3f2ed24cfdad4a7fea84a51d90

This also updates the htmx version because of a bug that our very own Ben Beecher fixed in the htmx library. I tripped it while working on this and included the upgrade


jake-kent (approved):
lgtm 🚀

gone:
replaced by other pr

PR #363, opened 2023-01-25
Add `pytest-dotenv`

samamorgan:
This adds `pytes-dotenv` and configures pytest to use .env or .env.example during tests.

This is really helpful for both local testing and CI. If a developer needs to, for instance, generate VCR cassettes, they will need live API keys for that and the local .env will be used. But when those recordings are repeated on CI, .env.example will be used instead, and the cassettes will be played back.


jake-kent (approved):
lgtm 🚀

PR #381, opened 2023-02-08
HYD-16 Frontend Validation

rwwiv:
This PR exposes the values of text + password inputs to parent elements by way of events tied to the input's id. 

It also includes an example of how you would use this on the frontend by adding a password match check to the signup form shown below. This example is pretty complicated and would be a lot simpler if we just wanted to show a form-level warning instead of modifying the input styling.

### DIFF
--- a/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/forms/password.ts
+++ b/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/forms/password.ts
@@ -1,23 +1,24 @@
 import AlpineInstance, { AlpineComponent } from "alpinejs";
 
-const password = (...args: unknown[]): AlpineComponent => ({
-  // Props
-  eventName: args[0] as string,
-  value: args[1] as string,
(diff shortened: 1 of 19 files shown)


{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/forms/password.ts
```diff
@@ -0,0 +1,23 @@
+import AlpineInstance, { AlpineComponent } from "alpinejs";
+
+const password = (...args: unknown[]): AlpineComponent => ({
```
jake-kent, line 3: It might be worth making the args a ts interface/type rather than unknown[] given that args[0] is eventName and args[1] is value both here and in the input.ts with the optional args[2] for type. Then you also wouldn't have to do `as string`. Maybe that would mess up other things but seems like something worth throwing in quickly

jake-kent (approved):
one small typescript typing suggestion but otherwise once the tests for this PR pass you're good to go

PR #387, opened 2023-02-13
Update several doc files

rwwiv:


jake-kent (approved):
A couple of notes on docs about the default .env file setup via the setup scripts, but otherwise looks good to me 🚀

PR #396, opened 2023-02-14
Text Area, Input Option, Multiple Input and Clearable File Widgets

jake-kent:


{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/forms/textarea.jinja
```diff
@@ -0,0 +1,21 @@
+{% from 'components/util.jinja' import attributes %}
+{% macro text_area(type="text", name=none, value=none, event_name="", color='primary', attrs={}, left_icon='', right_icon='') %}
+    <div x-data="input('update-{{ event_name | replace('_', '-') }}', '{{ value }}')"
+         class="w-full input">
+        <textarea
+            x-model.debounce.500ms="value"
+            {% set _ = attrs.update({"type": type, "name": name, "value": value}) %}
```
gone, line 7: might be a do statement?

jake-kent: okay I can try it as a do statement. there are probably also 100 other places where "set"s could be converted to "do"s

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/forms/textarea.jinja
```diff
@@ -0,0 +1,21 @@
...
+    </div>
+{% endmacro %}
+{# takes a django widget and calls our textarea macro with the appropriate args #}
+{% macro widget_to_textarea(widget) %}
+    {{ text_area(type=widget.type, name=widget.name, value=widget.value, event_name=widget.attrs.id, attrs=widget.attrs )}}
+{% endmacro %}
+
+{% if is_pattern_library %}
+    {{ text_area() }}
+{% endif %}
```
rwwiv, line 21: Should we remove this if we're not using pattern library yet?

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/components/util.jinja
```diff
@@ -1,7 +1,9 @@
 {% macro attributes(kwargs={}) %}
-    {% for name, value in kwargs.items() %}
-        {% if value %}
-            {{ name }}{% if value is not sameas True %}="{{ value }}"{% endif %}
-        {% endif %}
-    {% endfor %}
+    {% if kwargs != None %}
```
rwwiv, line 2: Nice add here. I think you could simplify it to `{% if kwargs %}` if you wanted?

rwwiv (changes requested):
I saw a couple of weird bits on these widgets:

## Textarea

Screenshot 2023-02-16 at 4 15 27 PM

The label is overlapped by the text. I think this should be animating like the regular text input?

## Clearable File

Screenshot 2023-02-16 at 4 16 08 PM

* Red: I think removing the label would be ideal for this guy
* Green: Likewise I think we shouldn't include a border for this input

---

The rest LGTM!

rwwiv (approved):
LGTM!

PR #399, opened 2023-02-17
Add date & datetime widgets

rwwiv:
This PR adds date and datetime pickers as the default date/datetime widgets for django.

---

## Date
Screenshot 2023-02-17 at 10 22 16 AM

## Datetime
Screenshot 2023-02-17 at 10 22 37 AM


samamorgan:
@rwwiv Have you checked how this renders in django admin?

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/forms/date.ts
```diff
@@ -0,0 +1,29 @@
+import AlpineInstance, { AlpineComponent } from "alpinejs";
+import flatpickr from "flatpickr";
+import { AlpineDataCallback } from "../../static_source/js";
+
+const date = (eventName: string, value: string): AlpineComponent => ({
```
samamorgan, line 5: I feel like date.ts and datetime.ts could be combined. Is there a good reason not to do that?

jake-kent: I think outside of the format string and timeenabled they could probably be combined into a single datetime.ts that's just called with a time=true/false arg. and even further I'd say the same about combining the macros and just adding a time=True/False arg based on the field.html that's calling them

rwwiv: My reasoning was to make customization easier, but in hindsight I can't picture someone wanting to make the two differ wildly in functionality or styling.

PR #443, opened 2023-03-27
Feature/update input styling

melissafelch:
# Short description

This pull request refactors the existing input CSS by using alpine state management to apply focus classes. This is primarily to address a growing dependence on the :has selector within recent PRs (specifically regarding validation), which broke most input functionality in firefox. Rather than relying on pseudoselectors, this PR implements alpine event listeners.

# Any point in the PR you think needs extra attention

There is (arguably) a bug with autofill fields where input text will overlap the label briefly if > 2 fields are filled at the same time. Once an autoselect option is selected, the label will adjust appropriately. (See video below for an example.) This can be resolved by using an [@transitionstart](https://developer.mozilla.org/en-US/docs/Web/API/Element/transitionstart_event) (or similar) event listener with Alpine, but current styling caused the event to fire multiple times. There is also the possibility that this event would ultimately be too sensitive for this usage, as it toggles whenever _any_ event is run on the attached input.

https://user-images.githubusercontent.com/80595928/228084578-fd6f0aac-f657-43cd-a74d-2a5e9f2d80ca.mov

# New Features or Fixes 
1. Toggle active class to input fields to remove reliance on pseudoselectors
2. Template in CSS to allow easier editing of colors and sizing
3. Switched sizing from px to rem to play nicer with tailwind  (with the exception of tom-select elements)
4. Improved previously jumpy animations
5. Added functional textareas
6. Removed unused class from checkboxes which caused odd rendering
7. Removed instances where multiple icons were displayed in inputs (primarily error states). In the case that an error indicator and functional icon (such as tooltip) were displayed, the error icon will retain the original's behavior.

# Video of feature working

Example demonstrating class toggling (in source panel); error state icon with proper functionality

https://user-images.githubusercontent.com/80595928/228086047-6c9fb58c-81a5-4384-a33a-1b413b35845a.mov

Example of a working textarea (note floating label)

https://user-images.githubusercontent.com/80595928/228086208-c8e9b04a-1770-4e01-82ae-f1058f07f02c.mov

### DIFF
--- a/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/static_source/css/base/_forms.scss
+++ b/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/static_source/css/base/_forms.scss
@@ -20,13 +20,8 @@ $border-color: var(--silver);
 $background-color: #fff;
 
 // label for focused input, Optionally can change the label color of focused element.
-@mixin label-focused($text-color:false) {
+@mixin label-focused {
   @apply top-0 left-0 pt-2 pb-2 text-xs leading-3;
-
-  @if $text-color {
-    color: $text-color;
-  }
-
   width: calc(100% - 0.75rem);
   background-color: $background-color;
 }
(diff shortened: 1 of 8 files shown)


brianedelman (approved):
Seems legit to me. Nice work!

PR #471, opened 2023-04-24
Updates

gone:


{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/config/urls.py
```diff
@@ -40,17 +40,20 @@
     ]
 
     if "robots" in settings.INSTALLED_APPS:
-        urlpatterns.append(re_path(r'^robots\.txt', include("robots.urls")),)
+        urlpatterns.append(
+            re_path(r"^robots\.txt", include("robots.urls")),
+        )
```
brianedelman, line 45: Would be nice to set it up to only include this for prod maybe or include a different one for dev

PR #280, opened 2022-11-30
Use cookiecutter organization name throughout project

samamorgan:
This moves the `org_name` capture up in priority during cookie cutting and uses that variable throughout the cut template in lieu of "Lightmatter".

PR #656, opened 2023-09-21
Feature/Use cookiecutter organization name throughout project

melissafelch:
# Short description

This pull request builds on work started in PR #280 to utilize org_name and other cookiecutter variables throughout the project. Because cookiecutter uses jinja templating to generate its files, we have created a unique string for blocks and variables which allows internal templating to be ignored as files are processed. This allows us to collect and insert information gathered from cookiecutter prompts.

# Any point in the PR you think needs extra attention
In order to use this, please be sure to update cookiecutter to at least v 2.3.0. Otherwise, things will break spectacularly!

I initially missed some files involved in migrations because they were omitted from vscode's search function. While I combed for any instances where cookiecutter was used, there is also a possibility that I may have missed perfectly valid jinja inside files which were not templated. I believe I caught them all but a critical eye would be appreciated!

# New Features or Fixes

1. Header and footers now utilize cookiecutter for project and organization
2. Default emails are now generated based on domain name, not organization name.
3. find & replace and find & replace and find & replace and find & replace and find & replace and find & replace and 

# Video of feature working

https://github.com/Lightmatter/django-hydra/assets/80595928/6604bd81-36e1-4d2b-9171-7ecb765c2485

Screenshot 2023-09-21 at 3 41 55 PM


hooks/pre_gen_project.py
```diff
@@ -1,10 +1,9 @@
 import re
 import sys
 
-
 MODULE_REGEX = r"^[a-z][_a-z]+$"
 
-module_name = "{{cookiecutter.repo_name}}"
+module_name = "{{ cookiecutter.repo_name }}"
```
melissafelch, line 6: weirdly enough, this file ignores the _jinja2_env_vars for strings?

gone:
Copier setup has a single variable now.
Hope you're doing good!

